PR #2, opened 2017-01-31
use explicit json.get<type>() to avoid compiler ambiguity errors

ronen:
Hi, thanks for making this library!

On my platform I was getting a handful of errors like this:

```
json-validator.cpp:677:31: error: use of overloaded operator '<' is ambiguous (with operand types 'std::size_t'
      (aka 'unsigned long') and 'const nlohmann::basic_json<std::map, std::vector, std::__1::basic_string<char>, bool, long long, unsigned long long,
      double, std::allocator, adl_serializer>')
                if (utf8_length( instance ) < attr.value()) {
```

which I fixed by using explicit `.get<type>()` to disambiguate the json values.

FYI here's the compiler info:
```
Apple LLVM version 8.0.0 (clang-800.0.42.1)
Target: x86_64-apple-darwin16.3.0
```
and I compile using the `-std=c++11` flag.

### DIFF
--- a/src/json-validator.cpp
+++ b/src/json-validator.cpp
@@ -315,7 +315,7 @@ void json_validator::validate(json &instance, const json &schema_, const std::st
 		if (ref == schema->end())
 			break;
 
-		auto it = schema_refs_.find(ref.value());
+		auto it = schema_refs_.find(ref.value().get<std::string>());
 
 		if (it == schema_refs_.end())
 			throw std::invalid_argument("schema reference " + ref.value().get<std::string>() + " not found. Make sure all schemas have been inserted before validation.");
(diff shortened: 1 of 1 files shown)


ronen:
...also one case of a native `bool` being compared against `json(bool)`  giving a similar error:

```
error: use of overloaded operator '==' is ambiguous (with operand types 'bool' and
      'json' (aka 'basic_json<>'))
                                if (ret.second == json(false))
```

fixed in a second commit

ronen:
> The change you're reverting here, will break it again.

> I agree, that comparing the == false should work, but it doesn't all the time.

To be clear, are you referring to this [line/(https://github.com/ronen/json-schema-validator/blob/0c454dcc1b3eb07dc0bc1028c1ec6bae11c42e91/src/json-validator.cpp#L459)?

```cxx
auto ret = array_to_set.insert(v);
if (ret.second == false)
```
the `ret.second` is really of type `bool` as per the [definition of std::set::insert](http://www.cplusplus.com/reference/set/set/insert/) --  it's not a `json` object.   How could testing a `bool` against `false` not work all the time?

The way it had been:

```cxx
if (ret.second == json(false))
```
was trying to test `bool` against `json`, which was failing for me in clang.

pboettch:
You're right. When fixing it in one place (with this commit) I also changed it here, which is wrong. Nicely spotted, thank you for insisting.

PR #3, opened 2017-01-31
Have validate() take a const json reference.

ronen:
In my app I have a `const json` object, which I wanted to validate.

The validator doesn't try to mutate the json document, so it was easy enough to add `const` to the declaration and propagate it as needed.

### DIFF
--- a/src/json-schema.hpp
+++ b/src/json-schema.hpp
@@ -164,10 +164,10 @@ class json_validator
 
 	std::map<json_uri, const json *> schema_refs_;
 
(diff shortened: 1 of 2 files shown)


pboettch:
Initially I had the idea of adding default values while validating the document. It will be done at some point but differently. Thanks for this change.

ronen:
Yes good point filling in defaults would change the document, I hadn't considered that.  But still nice to be able to validate a const document, when defaults aren't an issue.

Thanks for the speedy merge(s).

PR #4, opened 2017-01-31
Bug fix: Accept integer as a number in a list of types

ronen:
If the schema has

```
type: ["string", "number"]
```

and the corresponding value in the json document is an integer,  I was getting an error:

```integer is not any of ["string","number"] for ...```

Fixed in this PR.


pboettch:
There should be a test for it. But there isn't - I will create one.

ronen:
Yes you're right there should be a test.  Sorry, I should have made it.

PR #10, opened 2017-08-07
Build as shared library on Windows using Visual Studio compiler

gertvanhoey:
I have added preprocessor code to allow exporting the symbols from DLLs, conditionally disabled -Wextra and -Wall in case of Visual Studio compiler, and fixed two warnings.  After those changes, the only MSVC warning remaining is C4251 (https://msdn.microsoft.com/en-us/library/esew7y1w.aspx).  The subject of this issue is probably not a problem when building json-schema-validator and the code that uses it with the very same compiler.  Otherwise, it is flagging a potential problem, so this warning should probably not be suppressed.

### DIFF
--- a/app/json-schema-validate.cpp
+++ b/app/json-schema-validate.cpp
@@ -93,7 +93,7 @@ int main(int argc, char *argv[])
 	json document;
 
 	try {
-        std::cin >> document;
+		std::cin >> document;
 		validator.validate(document);
 	} catch (std::exception &e) {
 		std::cerr << "schema validation failed\n";
--- a/app/json-schema-validate.cpp
+++ b/app/json-schema-validate.cpp
@@ -93,7 +93,7 @@ int main(int argc, char *argv[])
 	json document;
 
 	try {
-		document << std::cin;
+        std::cin >> document;
 		validator.validate(document);
 	} catch (std::exception &e) {
 		std::cerr << "schema validation failed\n";
--- a/src/json-validator.cpp
+++ b/src/json-validator.cpp
@@ -344,7 +344,7 @@ void json_validator::validate(const json &instance, const json &schema_, const s
 		try {
 			validate(instance, attr.value(), name);
 			ok = false;
-		} catch (std::exception &e) {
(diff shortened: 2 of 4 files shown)


app/json-schema-validate.cpp
```diff
@@ -93,7 +93,7 @@ int main(int argc, char *argv[])
 	json document;
 
 	try {
-		document << std::cin;
+        std::cin >> document;
```
pboettch: Indentation should be a TAB (sorry ;-P )

src/json-validator.cpp
```diff
@@ -344,7 +344,7 @@ void json_validator::validate(const json &instance, const json &schema_, const s
 		try {
 			validate(instance, attr.value(), name);
 			ok = false;
-		} catch (std::exception &e) {
+        } catch (std::exception &) {
```
pboettch: Same here, indent.

pboettch:
Thanks for your contribution. Do you want me to make these minimalistic indent changes or can you update your PR?

gertvanhoey:
No problem I will fix this (my editor is set to using spaces and I didn't notice the inconsistency).  Thank you by the way for making your library available.  It works like a charm.

pboettch:
I squashed your 3 commits into one, makes more sense IMHO. You have to reset --hard your fork for future contributions. Thanks again.

gertvanhoey:
Dear Patrick,

Could you advise me how to do this reset –hard on the git command line ? –hard sounds to me like I could ruin the history of my fork if I don’t do it correctly.

Thanks,

Gert

pboettch:
Yes, be careful with `git reset --hard`. You can find good help with git everywhere over the internet.

What I wanted to say with 'I squashed your 3 commits' was that I create on my repository a new commit combining your 3 commits. From git's point of view this new commit has nothing to do with your 3 commits, so if you would merge, or rebase you will get conflicts. You need to reset your branch onto mine before doing any further commits (or you need to rebase your new commit on top of my master branch).

gertvanhoey:
Thank you

PR #13, opened 2017-11-20
Change error message for "minItems".

superamadeus:
The error message for "minItems" was the same as "maxItems". Change it to use the same verbiage as "minProperties".


pboettch:
Good catch, thanks.

PR #15, opened 2018-01-12
Added the option to disable dynamic library and build static instead

6mayank9:
It's useful for some platform (e.g. iOS) to build the validator as a static lib instead of dynamic.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,10 +26,17 @@ if(NOT TARGET json-hpp)
 endif()
 
 # and one for the validator
-add_library(json-schema-validator SHARED
-    src/json-schema-draft4.json.cpp
-    src/json-uri.cpp
-    src/json-validator.cpp)
+if (JSON_SCHEMA_VALIDATOR_DISABLE_SHARED)
+    add_library(json-schema-validator STATIC
+        src/json-schema-draft4.json.cpp
+        src/json-uri.cpp
+        src/json-validator.cpp)
+else()
+    add_library(json-schema-validator SHARED
+            src/json-schema-draft4.json.cpp
+            src/json-uri.cpp
+            src/json-validator.cpp)
+endif()
 
 target_include_directories(json-schema-validator
     PUBLIC
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,6 +36,9 @@ else()
             src/json-schema-draft4.json.cpp
             src/json-uri.cpp
             src/json-validator.cpp)
+    target_compile_definitions(json-schema-validator
+            PRIVATE
+            -DJSON_SCHEMA_VALIDATOR_SHARED_LIB)
 endif()
 
 target_include_directories(json-schema-validator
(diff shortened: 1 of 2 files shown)


pboettch:
Thanks for your pull-request. I clearly see the need for selecting the library-type. I haven't looked at it before too closely but I was surprised that I hard-coded the library-type to `SHARED`. The cmake-way of doing it is to do

```cmake
add_library(json-schema-validator <sources>)
```
so without specifying `STATIC` or `SHARED`. And then influence the library-type via the CMake-variable `BUILD_SHARED_LIBS`. This [cmake-help-page](https://cmake.org/cmake/help/v3.10/variable/BUILD_SHARED_LIBS.html) even says that adding this variable as an `option()` is "common practice".

Do you have the time to change the CMakeLists.txt to use the `BUILD_SHARED_LIBS`-variable?

6mayank9:
I'll do it once I get some time. Thanks for the review and the awesome project.

ianhbell:
As a counter-argument, it is also common practice to compile all the dependent libraries into static libraries and link them into a monolithic EXE (or shared library).  I think it would make more sense to have two possible targets, one for static, the other for shared.  That's what I usually do.

One wrinkle is to make sure that if you ARE going to make single DLL (shared library), that the -fPIC makes its way into all the dependent static libraries.  That can be done with the method in this answer: https://stackoverflow.com/a/38297422/1360263

pboettch:
Generating two different types libraries in one build is not the cmake-way I'm used to. Using the cmake-built-in variable `BUILD_SHARED_LIBS` to select whether a static or a dynamic library is built. It should be the user's choice: the initial call to cmake could look like this:

```
cmake -DBUILD_SHARED_LIBS=ON ../path/to/source
```

(Currently this is not working - hence this pull-request)

The same could be done with `set(CMAKE_POSITION_INDEPENDENT_CODE ON)`

I'll take a look.

pboettch:
I implemented your idea by using `BUILD_SHARED_LIBS`. Feel free to reopen if it does not work for you.

ianhbell:
The use of BUILD_SHARED_LIBS does not solve the problem.  Let's suppose you have a library A, that has as a dependency your library, as well as 6 other libraries (actually, my use case).  If you pass BUILD_SHARED_LIBS, then all 6 libraries will be built as shared libraries and linked together, but that brings with it ALL kinds of challenging problems (google "dll hell" for more information).  For that reason, it is nice to build all the dependent libraries into static libraries and link them into the top-level object.  Goodbye DLL hell.

I currently do not use any of your cmake targets and I roll my own to have full control, but all I'll say is that BUILD_SHARED_LIBS is not the right solution.

pboettch:
Regarding `BUILD_SHARED_LIBS, it can be handled on a per-subdir basis if needed.

Regarding building two libraries, I must say sorry, I tend to disagree.

When starting to build shared and static libraries in one build, why not also building a version with Debug and Release as well? No, that would be too much.

What I want to say is that it should be up to the user to choose how he wants to build the library. As static or not, with debug-info or not. The project just needs to make sure that all possibilities work. 

I personally prefer a library which offer me one version but is correctly configurable than a library which build things I don't use.

It's a pity that dynamic libs on Windows platforms seems to be 'hell'.

Issue #18, opened 2018-02-24
Runnable example on main page


ibell:
I was able to get your library running (so far it's great!), but expect a few PR forthcoming.  Especially related to the documentation and cmake integration, both of which could use a wee spot of love.

For the sake of new users, it would be nice if your example on the main page was runnable.  Like copy-paste runnable.  Deciphering how to get it working otherwise is a little spot of bother, but not too bad.

PR #19, opened 2018-02-24
Add a runnable and useful example

ibell:
Make the example in the README.md more useful because it can be run directly and demonstrates some useful features of the library.

Removes the loader callback, which was VERY unclear to new users and doesn't seem strictly required.

See also #18


pboettch:
Great! Thank you. The more you work with it, the more you are getting blind for new users and potential show-stoppers.

PR #20, opened 2018-02-26
Use more explicit path to nlohmann/json.hpp

ianhbell:
The path "json.hpp" is a bit dangerous because that path could be found in many libraries in theory and include directory determination could get confused by finding the wrong one.

### DIFF
--- a/src/json-schema.hpp
+++ b/src/json-schema.hpp
@@ -36,7 +36,7 @@
 #    define JSON_SCHEMA_VALIDATOR_API
 #endif
 
-#include <json.hpp>
+#include <nlohmann/json.hpp>
 
 // make yourself a home - welcome to nlohmann's namespace
 namespace nlohmann


pboettch:
My implementation also changes the CMakeLists.txt-file. What do you think? Feel free to re-open. Thanks for triggering.

ianhbell:
It is better to make the change in the header itself and then have the include path in cmake.  The main reason is because of potential for name clashes.  If you have a large project with many moving pieces, it becomes more and more difficult to find header names that are actually unique and to include the header you actually wanted.

This follows the general rule: explicit is always better than implicit. 

I've learned this lesson the hard way, and now in every project have an include folder, inside which I have a project folder, and inside that, the header.  Then you add the include path to the include directory, and in all files, so something like:

``` c++
#include "library/a_header.hpp"
```

Niels Lohmann does the same thing, for what it's worth, in his json library.  Imagine instead your header was called "math.h".  Think of all the fun include path name clashes you would have.

pboettch:
That's what I did in `json-schema.hpp`. Have you not seen it?

This additionally required changes to the CMakeLists.txt. That's all I tried to say.

ianhbell:
_Mea culpa_ - I didn't see that change.  If you are going to make the change anyway, it's good practice on github to accept the PR and then have as a supplement your commit modifying the CMakeLists.txt

pboettch:
I thought about that, but that would have created a commit which would have broken the build.

PR #21, opened 2018-02-28
Add patch to fix dllimport preprocessor on WIN32

ibell:


ibell:
I was lazy and did this PR in the web interface- you'll probably want to squash-merge my PR to remove my merge commit from this history

PR #23, opened 2018-03-15
Updated CMakeLists.txt to add install target when needed.

anthonytw:
As it stands, the install target for the header will never be created because it's predicated on the existence of the `json-hpp` target. But if the target does not exist initially it's created before hitting that conditional, so it's always false.

This change sets a flag if the install target wasn't initially defined upon executing the script, and if not, it will create the install target. So now you can install as you'd expect:

```
cmake -D NLOHMANN_JSON_DIR=<...> -D CMAKE_INSTALL_PREFIX=<...> ..
make install
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,7 @@ cmake_minimum_required(VERSION 3.2)
 # and associate the path to json.hpp via target_include_directories()
 if(NOT TARGET json-hpp)
     set(NLOHMANN_JSON_DIR "" CACHE STRING "path to json.hpp")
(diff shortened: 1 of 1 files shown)


pboettch:
Why closed? I think you're right, there is a problem. Why is the intsall-rule only created in the json-hpp target is not set. The install-rule could always be there, couldn't it?

anthonytw:
Sorry, I was going to create another push request, I realized that this install doesn't actually add the library or cmake configuration files. I'll add another pull request with everything. :-)

PR #24, opened 2018-03-15
Updated install scripts

anthonytw:
This fixes the issue that install scripts are never generated. They're still only generated if the `json-hpp` target doesn't initially exist, but that can be another modification if you want it.

This also adds the library to the install targets and it cobbles together some code to generate the CMake configuration files (so you can used find_package(json-schema-validator)). DISCLAIMER: This is the first time I've set that up, so it might not be 100% correct, but I was able to successfully import it into another project and link it through CMake's find_package module so I think it's working.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,8 @@ project(json-schema-validator CXX)
 
 cmake_minimum_required(VERSION 3.2)
 
+set(PROJECT_VERSION 1.0.0)
+
 # if used as a subdirectory just define a json-hpp-target as add_library(json-hpp INTERFACE)
 # and associate the path to json.hpp via target_include_directories()
 if(NOT TARGET json-hpp)
@@ -22,21 +24,16 @@ if(NOT TARGET json-hpp)
     endif()
 
     # create an interface-library for simple cmake-linking
+    include_directories(${NLOHMANN_JSON_REALPATH})
     add_library(json-hpp INTERFACE)
-    target_include_directories(json-hpp
-        INTERFACE
-            ${NLOHMANN_JSON_REALPATH})
 endif()
 
 # and one for the validator
-add_library(json-schema-validator
+include_directories(src)
+add_library(${PROJECT_NAME}
     src/json-schema-draft4.json.cpp
     src/json-uri.cpp
     src/json-validator.cpp)
-
-target_include_directories(json-schema-validator
-    PUBLIC
-        ${CMAKE_CURRENT_SOURCE_DIR}/src)
 
 target_compile_features(json-schema-validator
     PUBLIC
@@ -47,9 +44,6 @@ if(NOT MSVC)
         PUBLIC
             -Wall -Wextra)
 endif()
-target_link_libraries(json-schema-validator
-    PUBLIC
-        json-hpp)
 if(BUILD_SHARED_LIBS)
     target_compile_definitions(json-schema-validator
         PRIVATE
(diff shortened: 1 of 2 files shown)


anthonytw:
Will this make it in to the repo? It is useful to have.

pboettch:
If everything works as expected, I will check it (in) tomorrow.

CMakeLists.txt
```diff
@@ -21,22 +24,17 @@ if(NOT TARGET json-hpp)
     endif()
 
     # create an interface-library for simple cmake-linking
+    include_directories(${NLOHMANN_JSON_REALPATH})
```
pboettch: Using `include_directories()` is discouraged since some time. Only in very rare cases this should be used. Did you remove the `target_xxx()`-stuff because it does not allow you to create the package-files with export?

anthonytw: I believe it caused problems on the import. Since writing this and trying to incorporate these cmake build scripts in other projects I *think* I now understand it to be the case that if your cmake project relies on other projects (that you include via find_package, say) any project that includes yours must also include the dependencies explicitly, which was sort of non-intuitive. I'm not sure how to get around that though.

CMakeLists.txt
```diff
@@ -71,13 +66,47 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
...
         FILES
-           ${CMAKE_CURRENT_SOURCE_DIR}/src/json-schema.hpp
-        DESTINATION
-            ${CMAKE_INSTALL_PREFIX}/include
-    )
+            "${CMAKE_CURRENT_BINARY_DIR}/cmake-config/${PROJECT_NAME}Config.cmake"
+            "${CMAKE_CURRENT_BINARY_DIR}/cmake-config/${PROJECT_NAME}ConfigVersion.cmake"
+        DESTINATION ${ConfigPackageLocation}
+        COMPONENT Devel)
 endif()
```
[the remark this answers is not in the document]
anthonytw: Probably. I pulled this from cmake's documentation. Since this, I've implemented these imports a few more times, and I've found some of these commands seem a bit redundant or unnecessary, but from the documentation it's not clear to me what's actually required and what's not, and what's expected practice. For such a common function it's horribly documented (which is probably why every cmake package has it's own uniquely frustrating and opaque import semantics).

anthonytw:
Agreed with your concerns, this probably isn't implemented as it should be, but given cmake documentation and posts online it's not clear at all what the recommended practice should be to define these importable modules. But this is a really helpful functionality for incorporating this package in other projects, maybe it just needs to be reworked.

pboettch:
Yeah, that's what is a big problem with CMake, but also a blessing. By not forcing you to do everything in a certain way, you a forced to find a way which fits your project.

Could you describe what you would like to do with this library's CMake from a pure user point of view? With commands you are going to issue to achieve it (might be as simple and `make` and `make install` and user's CMakeLists.txt (`find_package`?) ? If you do `find_package()`, where does the `json.hpp` will come from?

anthonytw:
From a user's point of view, the end goal is to pull in the package dependencies automatically in cmake using `find_package(json-schema-validator)`. Ideally it wouldn't require me to explicitly include `find_package(json)` but it's not the end of the world if it's needed, although if it's possible to generate those configuration scripts such that they attempt to import that dependency automatically that would be optimal. So from this package, being able to run `make install` which should move the appropriate dependencies and generate the CMake configuration scripts is all that's needed.

CMakeLists.txt
mxmlnkn: I think [`INTERFACE_LINK_LIBRARIES`](https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_LINK_LIBRARIES.html) allows you to implicitly import dependencies.
So, something like (untested!)
set_property(TARGET json-schema-validator PROPERTY INTERFACE_LINK_LIBRARIES json-hpp)

Imo @pboettch is right and using `include_directories` instead of `target_include_directories` is very bad style and could lead to more problems for larger projects and therefore shouldn't be merged like this.

pboettch:
PR #73 is an updated version of this PR. Closing it.

Issue #29, opened 2018-03-30
exclusiveMaximum and exclusiveMinimum behavior inverted


RPGillespie6:
According to JSON Schema spec,

exclusiveMaximum/Minimum is a flag that tells you if the maximum value should be included or not.

This library has 2 bugs:

1. The behavior in `validate_numeric` is inverted (i.e. if exclusiveMaximum is present, it does an inclusive comparison)
2. The value of the flag is not checked.

The correct behavior would be to do something like:

```c++
    const auto &excl = schema.find("exclusiveMaximum");
    bool exclusive = (excl != schema.end()) ? excl.value().get<bool>() : false;
```

And then in the comparison logic:
```c++
    if (exclusive)
        return value > max; //exclusive comparison
    else
        return value >= max; //inclusive comparison
```

This is assuming you have a type-independent comparison utility:

```c++
//uses the types the nlohmann uses under the hood instead of constricting to double
bool violates_numeric_maximum(const json &max, const json &value, bool exclusive)
{
    if (exclusive)
        return value > max;
    else
        return value >= max;
}
```

I can prepare a PR

pboettch:
Please do a PR. :+1:

The scary thing is that the JSON-Schema-Test-suite succeeds the exclusive Min/Max-tests. Because of the impression of double. (it compares 1.1 and 1.1). Double seems not to be able to represent that value exactly.

pboettch:
I was wrong. There is currently no bug in this library. But I agree the folded logic (violate vs. exclusive vs. accept) makes it hard to understand.

pboettch:
If you agree, please close this bug.

RPGillespie6:
Ah, my mistake. You are correct.

In that case the only bug was that the library was previously not examining the boolean value of "exclusiveMaximum" as far as I could tell.

pboettch:
I already closed this issue via a commit-message (which has disappeared in the meantime). Please feel free to comment the last commit and create new PR for anything you'd like.

PR #30, opened 2018-04-02
Fix #28 and #29

### LINKED ISSUE #28: Use schema+nlohmann type to determine numeric validation type
[Nlohmann currently provides an API](https://nlohmann.github.io/json/classnlohmann_1_1basic__json_a2b2d781d7f2a4ee41bc0016e931cadf7.html) that returns one of 3 different numeric types:

```
number (integer) | value_t::number_integer
number (unsigned integer) | value_t::number_unsigned
number (floating-point) | value_t::number_float
```

Consider the following:

```c++
#include <iostream>

using namespace std;

int main() {
  uint64_t x = 99999999999999997;
  uint64_t y = 99999999999999999;
  double a = x;
  double b = y;
  cout << to_string(x == y) << " " << to_string(a == b) << endl;
}
```
In this case the library will falsely validate numbers since there isn't enough fractional bits in the double to distinguish the two numbers.

This library should:

1. Check if the Schema v4 type is `number` vs. `integer`

a. If the Schema specifies `integer`, then it should first check that the provided number is either `value_t::number_integer` or `value_t::number_unsigned`

If not, the validation should fail.

b. When doing the comparison, the type that allows the maximum precision should be chosen.

For example, if the schema type is `integer`, then, `int64_t` or `uint64_t` should be used. If the comparison value and actual value are both `value_t::number_unsigned`, then choose `uint64_t`, otherwise choose `int64_t`.

Otherwise, if schema type is `number`, `double` should be used.

2. Mention in the README Weaknesses section that due to c++ constraints, this library can only validate against 64-bit integers, unsigned integers, or double precision even though the JSON spec has arbitrary precision (i.e. `bignum`)

RPGillespie6:
Made `validate_numeric` a template function so it can handle one of 3 numeric types: `int64_t`, `uint64_t`, or `double`. Also corrected exclusive minimum/maximum comparison bug.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,12 @@ lives in his namespace.
 Schema-reference resolution is not recursivity-proven: If there is a nested
 cross-schema reference, it will not stop.  (Though I haven't tested it)
 
+Numerical validation uses `int64_t`, `uint64_t` or `double`, depending on if the schema type is "integer" or "number". Bignum (i.e. arbitrary precision and range) is not supported at this time.
+
+Unsigned integer validation will only take place if the following two conditions are true:
+- The nlohmann `type()` of the json object under validation is `nlohmann::json::value_t::number_unsigned`
+- The schema specifies a numerical minimum greater than or equal to 0
+
 # How to use
 
 The current state of the build-system needs at least version **3.1.1** of NLohmann's
--- a/src/json-validator.cpp
+++ b/src/json-validator.cpp
@@ -186,64 +186,98 @@ void validate_boolean(const json & /*instance*/, const json &schema, const std::
 	validate_type(schema, "boolean", name);
 }
 
-void validate_numeric(const json &schema, const std::string &name, double value)
+template <class T>
+bool violates_numeric_maximum(T max, T value, bool exclusive)
 {
+	if (exclusive)
+		return value > max;
+
+	return value >= max;
+}
+
+template <class T>
+bool violates_numeric_minimum(T min, T value, bool exclusive)
+{
+	if (exclusive)
+		return value < min;
+
+	return value <= min;
+}
+
+template <class T>
+void validate_numeric(const json &instance, const json &schema, const std::string &name)
+{
+	T value = instance;
+
 	// multipleOf - if the rest of the division is 0 -> OK
 	const auto &multipleOf = schema.find("multipleOf");
 	if (multipleOf != schema.end()) {
-		if (multipleOf.value().get<double>() != 0.0) {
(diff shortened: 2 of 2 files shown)


RPGillespie6:
This is still a work in progress, but I wanted you to take a look at what I've done so far.

Need to still update the unit tests.

README.md
```diff
@@ -34,6 +34,12 @@ lives in his namespace.
 Schema-reference resolution is not recursivity-proven: If there is a nested
 cross-schema reference, it will not stop.  (Though I haven't tested it)
 
+Numerical validation uses `int64_t`, `uint64_t` or `double`, depending on if the schema type is "integer" or "number". Bignum (i.e. arbitrary precision and rage) is not supported at this time.
```
pboettch: Lots of rage with these numbers types, but I think you want to say range here.

src/json-validator.cpp
```diff
@@ -186,64 +186,98 @@ void validate_boolean(const json & /*instance*/, const json &schema, const std::
...
+bool violates_numeric_minimum(T min, T value, bool exclusive)
+{
+	if (exclusive)
+		return value < min;
+
+	return value <= min;
+}
+
+template <class T>
+void validate_numeric(const json &instance, const json &schema, const std::string &name)
```
RPGillespie6: Wondering if I should do a `static_assert` to prevent usage of this template with types other than `int64_t`, `uint64_t`, or `double`. Probably ok since it's a non-public function.

src/json-validator.cpp
```diff
@@ -186,64 +186,98 @@ void validate_boolean(const json & /*instance*/, const json &schema, const std::
...
+
+template <class T>
+bool violates_numeric_minimum(T min, T value, bool exclusive)
+{
+	if (exclusive)
+		return value < min;
+
+	return value <= min;
+}
+
```
pboettch: These two functions are wrong actually regarding the exclusive argument: If exclusive is true, the value violates the schema if it is inclusive regarding the min-value. So somewhere things need to be swapped. Either `violates` becomes `accept` or the `if` branches are swapped.

pboettch:
I close this PR, I integrated the changes. I fixed the exclusive-checks and improved the multiple-of-check compared to the original pull-request. Authorship was maintained.

src/json-validator.cpp
RPGillespie6: Ah, you are correct. There was not a bug in the comparison logic in the first place, just that it wasn't examining whether the field was true or false.

PR #32, opened 2018-05-11
Add CMake installation and selective build option support

RPGillespie6:
Adds support for the following CMake options:
- BUILD_TESTS
- BUILD_EXAMPLES
- CMAKE_INSTALL_PREFIX


RPGillespie6:
Not sure why TravisCI is failing... cannot reproduce locally?

pboettch:
Travis log says:

```
  install TARGETS given no ARCHIVE DESTINATION for static library target
  "json-schema-validator".
```

RPGillespie6:
Ah, I didn't test the static library install case. Should be fixed now.

PR #33, opened 2018-06-06
resolve "$ref" with relative uri correctly. fixes #9

### LINKED ISSUE #9: URI paths are not resolved correctly
In the current version of the library URI paths are not resolved correctly.

The problem is the handling of the (non-)presence of trailing slashes of the current path of an URI when trying to add or replace it with a new one.

The recursive-schema-test from the JSON-schema-examples triggers this problem.

simonlo-sony:
As noted in #9, resolving of relative references is broken.

Some examples resolved by this PR:

Given a current id of "foo.json", a relative reference "bar.json" should resolve to "bar.json".
Given a current id of "foo/bar.json", a relative reference "baz.json" should resolve to "foo/baz.json".
Given a current id of "foo/bar.json", a relative reference "baz/qux.json" should resolve to "foo/baz/qux.json".

(Relative URIs with paths containing dot-segments (".." or ".") will work if the schema loader handles them, but they won't be removed in the schema refs, so these schemas may still be loaded more than once.)

This is achieved by fixing the ``json_uri::from_string`` handling of relative paths, and making sure the ``resolver`` is only applied to the schema object once (because it modifies the "$ref" values in-place).

### DIFF
--- a/src/json-validator.cpp
+++ b/src/json-validator.cpp
@@ -305,10 +305,10 @@ void json_validator::insert_schema(const json &input, const json_uri &id)
 	// allocate create a copy for later storage - if resolving reference works
 	std::shared_ptr<json> schema = std::make_shared<json>(input);
 
-	do {
-		// resolve all local schemas and references
-		resolver r(*schema, id);
+	// resolve all local schemas and local and external references
+	resolver r(*schema, id);
 
+	do {
 		// check whether all undefined schema references can be resolved with existing ones
 		std::set<json_uri> undefined;
 		for (auto &ref : r.undefined_refs)
--- a/src/json-uri.cpp
+++ b/src/json-uri.cpp
@@ -86,8 +86,10 @@ void json_uri::from_string(const std::string &uri)
 		auto path = url.substr(pos);
 		if (path[0] == '/') // if it starts with a / it is root-path
 			path_ = path;
-		else // otherwise it is a subfolder
+		else { // otherwise it is a relative-path
+			path_ = path_.substr(0, path_.find_last_of('/') + 1);
 			path_.append(path);
+		}
 
 		pointer_ = local_json_pointer("");
 	}


simonlo-sony:
JOSN-Suite::ref now passes.

JSON-Suite::refRemote still fails. This is a different issue - the test case "base URI change - change folder" fails because it requires subschemas to be addressable via multiple URIs. See http://json-schema.org/latest/json-schema-core.html#rfc.section.8.2.2.

pboettch:
Could you check the changes I made on https://github.com/pboettch/json-schema-validator/tree/fix-issue-9 it resolves the issue you fixed and the multiple URI/schema. Making ref and refRemote-tests work.

src/json-validator.cpp
```diff
@@ -305,10 +305,10 @@ void json_validator::insert_schema(const json &input, const json_uri &id)
 	// allocate create a copy for later storage - if resolving reference works
 	std::shared_ptr<json> schema = std::make_shared<json>(input);
 
-	do {
-		// resolve all local schemas and references
-		resolver r(*schema, id);
+	// resolve all local schemas and local and external references
+	resolver r(*schema, id);
```
pboettch: Hmm, putting resolver creation outside the loop. I'm not sure whether this is a good idea concerning another round of ref-resolving when some new external refs have been updated.

simonlo-sony: I have tested  https://github.com/pboettch/json-schema-validator/tree/fix-issue-9, unfortunately resolving of relative references is still broken.  The casue of issues are:
1. current path (path_) can be emptied, while  json_uri::update() is being called. i.e. json-uri.cpp line 103 `path_.back()` will cause a crash.
2. each time when resolver is being called, the current path of the external ref is updated, having resolver doing another round of ref-resolving, would end up constructing an invalid path for the external ref. 

Please find attached schema set, which could be easily be showing the above issues. 
[schema.zip](https://github.com/pboettch/json-schema-validator/files/2089785/schema.zip)

Test was done using json-schema-validate to load the base.json schem, and schema loader to load the other relative schema.

```
static void loader_local(const json_uri &uri, json &schema)
{
    	const std::string base("D:/DATA/tmp/schema/"); // where schema set are installed 
	std::fstream lf(base + uri.path());
	if (!lf.good())
		throw std::invalid_argument("could not open " + uri.url() + " tried with " + uri.path());
	try {
		lf >> schema;
	}
	catch (std::exception &e) {
		throw e;
	}
}
```

pboettch: Thanks for your valuable feedback, I will take a look into it.

pboettch: You're right, your version of resolving the relative path is working correctly. But there are still some issues with the ref and refRemote tests.

pboettch:
A new version is almost ready, solving this issue correctly.

PR #39, opened 2018-11-22
fixed compile error error: error: aggregate ‘std::stringstream s’ has incomplete type and cannot be defined    std::stringstream s

mabondarenko:


src/json-uri.cpp
```diff
@@ -24,164 +24,165 @@
...
 #include "json-schema.hpp"
+#include <sstream>
 
 namespace nlohmann
 {
 
 void local_json_pointer::from_string(const std::string &r)
 {
-	str_ = "#";
+  str_ = "#";
```
pboettch: Could you re-create a commit (and push-force) without all these white-space-changes?

mabondarenko: sure. already done.

PR #46, opened 2019-01-21
first experiment with upgrading from v1.0.0

garethsb-sony:
Pleased to see the rewrite, as we've found things like repeated [recompiling of regex patterns in v1.0.0](https://github.com/pboettch/json-schema-validator/blob/1.0.0/src/json-validator.cpp#L694-L700), etc. to be a potential performance problem.

I've only just started playing with the rewrite, but I thought I'd immediately track things I found while trying to upgrade.

Four commits so far:

1. One missing #include
2. Backwards incompatibility - json_validator used to be copyable, it now wasn't even movable, so this commit fixes that
3. Delete whitespace either side of the '#' which indicates fragment identifier between the URI and json-pointer or local name (i.e. from a $id)
4. Update test for 3.

I'm seeing an exception from within test/id-ref.cpp:

```
  [json.exception.parse_error.107] parse error at byte 1: JSON pointer must be
  empty or begin with '/' - was: 'foo'
```

I haven't yet investigatd further, but does that mean $ref using a local name isn't working currently?

For reference, spec quote:

> All fragment identifiers that do
> not match the JSON Pointer syntax MUST be interpreted as
> plain name fragment identifiers.

### DIFF
--- a/src/json-validator.cpp
+++ b/src/json-validator.cpp
@@ -591,8 +591,8 @@ class numeric : public schema
 
 	std::pair<bool, json::number_float_t> multipleOf_{false, 0};
 
-	// multipleOf - if the rest of the division is 0 -> OK
-	bool violates_multiple_of(json::number_float_t x) const
+	// multipleOf - if the remainder of the division is 0 -> OK
+	bool violates_multiple_of(T x) const
 	{
 		json::number_integer_t n = static_cast<json::number_integer_t>(x / multipleOf_.second);
 		double res = (x - n * multipleOf_.second);
--- a/src/json-validator.cpp
+++ b/src/json-validator.cpp
@@ -267,27 +267,20 @@ class logical_combination : public schema
 		for (auto &s : subschemata_) {
 			basic_error_handler err;
 			s->validate(instance, err);
-
-			if (err) {
-				//sub_schema_err << "  one schema failed because: " << e.what() << "\n";
-				if (combine_logic == allOf) {
-					e.error("", instance, "at least one schema has failed, but ALLOF them are required to validate.");
-					return;
-				}
-			} else
+			if (!err)
 				count++;
 
-			if (combine_logic == oneOf && count > 1) {
-				e.error("", instance, "more than one schema has succeeded, but only ONEOF them is required to validate.");
-				return;
-			}
-			if (combine_logic == anyOf && count == 1)
+			if (is_validate_complete(instance, e, err, count))
 				return;
 		}
 
-		if ((combine_logic == anyOf || combine_logic == oneOf) && count == 0)
+		if (count == 0)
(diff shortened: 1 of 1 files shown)


pboettch:
Thank you so much for contributing.

Could you make an issue for the empty json-pointer, please? With an example which fails but should not.

src/json-uri.cpp
```diff
@@ -114,7 +114,7 @@ std::string json_uri::to_string() const
 {
 	std::stringstream s;
 
-	s << location() << " # " << pointer_.to_string();
+	s << location() << "#" << pointer_.to_string();
```
pboettch: I explicitly used spaces here, the to_string()-function should only be used for debugging (and unit-testing). At one point I had it even removed.

Or do you see a real usage for this method or the json-uri-class? 

The `json_uri`-class is really very basic and should be used extensively. Except for getting the components in the loader-callback.

Maybe it should go completely private and the loader-callback receives just a simple structure containing the fields... I don't know.

pboettch:
I will cherry-pick the interesting, no question commits from this PR right now. Thanks again.

src/json-uri.cpp
garethsb-sony: Ok, I wondered if that was the intent. I am still a bit unclear whether the schema loader callback will only ever get URIs which don't have a fragment, and your code will take care of multiple json-pointer references into those schemas? If so, it would clarify things if the argument to the loader callback couldn't have a fragment identifier.

garethsb-sony:
Thanks, @pboettch. I've reset the PR branch before the "#"-related commits, and will continue adding to it.

src/json-uri.cpp
pboettch: Yes, schema-references should work as you say: during schema insertion a sub-schema can be referenced by different URIs. That was _the_ reason for making a new version, because impossible to resolve with 1.0.0.

garethsb-sony: Hi @pboettch,

Sorry, that wasn't my point.

If the schema validator encounters a reference to "http://example.com/schema.json#/definitions/A", should the URL passed to the loader be just "http://example.com/schema.json" or should it have the JSON pointer?

If the schema validator subsequently encounters a reference to "http://example.com/schema.json#/definitions/B", will the loader be called again, or will the JSON pointer "/definitions/B" already have been resolved after the previous load?

Thanks!

garethsb-sony:
Recent commits fix one bug and mean that src/json-validator.cpp now compiles without any warnings in my environment: Visual Studio 2015 with ``/W4 /wd"4100"`` (i.e. Level4, with "warning C4100: '\<variable\>': unreferenced formal parameter" disabled)

src/json-uri.cpp
pboettch: Sorry for my misunderstanding. The loader will only be called once per file.

PR #49, opened 2019-02-01
Use proposed json_pointer::operator/

garethsb-sony:
from from https://github.com/nlohmann/json/pull/1469
and very small tweaks to a few error messages as discussed in issue #45

PR #52, opened 2019-02-05
Error reporting improvements

garethsb-sony:
Based on top of PR #49 

* report first error message of first subschema failure for 'allOf'
* make other error messages more consistent
* json_validator::validate can be const
* introduce error_handler 'interface'
* typedefs for schema_loader and format_checker functions
* remove some commented-out debugging code


garethsb-sony:
Now, in order to improve the reporting of the messages for 'anyOf' and 'oneOf' where no subschema is successful, it would be necessary to accumulate subschema failure details, but it's not currently clear how the implementation should select which subschema failure to report, or how to report multiple such failures.

*Either* ``logical_combination`` could format any/all subschema failures into a string message, which means the user-supplied ``error_handler`` has no control (and e.g. cannot access the ``json_pointer`` for the subschema failures), *or* we have to make the ``error_handler`` interface more complicated (such as passing a list of failures).

pboettch:
Your changes look good.

Regarding the heuristic, it could done in a custom error handler, in a separate file as this is not necessarily the core part of this library. 

In the json-schema-group they are discussing the standardization of error reporting. This might solve the issue of every schema-validator find their own solutions.

garethsb-sony:
> In the json-schema-group they are discussing the standardization of error reporting. This might solve the issue of every schema-validator find their own solutions.

OK, interesting.

At the moment, I think the only way to solve this generally is to make the ``error_handler`` callback be something like:

```C++
void error(const json_pointer& ptr, const std::string& message, const json& details);
```

where ``details`` may be null, or in the case of a logical-combination schema, an array of sub-schema errors (with their own details):

```json
[
  {
    "ptr": "/foo/bar",
    "message": "...",
    "details": [
      {
        "ptr": "/foo/bar/baz",
        "message": "..."
      },
      {
        "ptr": "/foo/bar/qux",
        "message": "...",
        "details": [
          // etc.
        ]
      }
    ]
  }
]
```

This would allow accumulating the details (i.e. the reasons why all sub-schemas failed) for 'anyOf' before calling the user ``error_handler``.

In the case of 'allOf', if the user were going to throw an exception to stop parsing, they would rather get a callback on the *first* failure. On the other hand, a user who wants the full error information may rather have a single error callback, including the accumulated errors of all sub-schemas.

In the case of 'oneOf', a similar thing applies. A user who wants to throw an exception quickly would like a callback on the *second* successful sub-schema, whereas a user who wants full information may rather have a callback once *all* the sub-schemas have been attempted. In both cases, the details should include the json-pointer of the multiple schemas that have been successful thus causing 'oneOf' validation to fail.

We could support configuration of 'fail fast' vs. 'complete details' if necessary.

pboettch:
I tried to rebase your changes manually, but I failed. There are severals things mixed within one commit. Maybe it would help if these things would be separated.

garethsb-sony:
I have rebased the changes onto master.

As a reminder these changes are some initial improvements:

* report first error message of first subschema failure for 'allOf'
* make other error messages more consistent
* json_validator::validate can be const
* introduce error_handler 'interface'
* typedefs for schema_loader and format_checker functions
* remove some commented-out debugging code

The only way to get really good error messages for schemas that use logical combinations of subschemas, is to do something like I propose above (https://github.com/pboettch/json-schema-validator/pull/52#issuecomment-461781277). We can discuss that in a new issue if you prefer?

pboettch:
I integrated your nice changes "manually" by keeping your authorship - because you merging master in your branche made it a git-mess. IMO, It's always better to rebase and push-force to PR when upstream changes have been made.

Thanks.

PR #59, opened 2019-06-12
Open files in read-only mode in validator

szotsaki:
Some build and test systems offer only read-only file system, so open JSON and schema files in read-only mode (since we anyhow don't want to write into them).

(Additionally, catch exceptions with `const` qualifiers, if I was already there.)


pboettch:
Cool, thanks for contributing.

PR #66, opened 2019-07-22
Added support to export cmake targets

Bergmann89:
I use the export functionality of cmake in my projects. If I use the json-schema-validator in my projects, cmake prints the following error.

`CMake Error: install(EXPORT "my-project" ...) includes target "my-project-static" which requires target "json-schema-validator" that is not in the export set.`

To fix this, the json-schema-validator must export it's targets too. The pull request fixes the error I have with my projects, but I think there must be added something more to fully support the cmake exports.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,6 +120,6 @@ if (BUILD_TESTS)
     add_subdirectory(test)
 endif()
 
-install ( EXPORT json-schema-validator
-          NAMESPACE json-schema-validator::
-          DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}" )
+install(EXPORT json-schema-validator
+        NAMESPACE json-schema-validator::
+        DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -117,3 +119,7 @@ if (BUILD_TESTS)
     enable_testing()
     add_subdirectory(test)
 endif()
+
+install ( EXPORT json-schema-validator
```
pboettch:
```suggestion
install(EXPORT json-schema-validator
```

CMakeLists.txt
```diff
@@ -117,3 +119,7 @@ if (BUILD_TESTS)
...
+          NAMESPACE json-schema-validator::
```
pboettch:
```suggestion
        NAMESPACE json-schema-validator::
```

schaumb:
I tried to use this patch with [Hunter](https://github.com/ruslo/hunter), but it's says:

`CMake Error: install(EXPORT "json-schema-validator" ...) includes target "json-schema-validator" which requires target "json-hpp" that is not in the export set.`

pboettch:
Please feel free to check with #73 which is an updated version of #71 and a more modern version of yours.

PR #67, opened 2019-08-15
[DRAFT] Provide richer error info for logical combination schemas

garethsb:
Follow-up to #52.

If a schema includes an "allOf", "anyOf" or "oneOf" near to the root, error messages are almost useless unless they go beyond "one of the subschemas is required to validate" and identify what the failures in each of the subschemas were. Gathering this error info would allow heuristics to be applied to identify the 'best match' (see https://github.com/json-schema-org/json-schema-spec/issues/632), or most nearly matching subschema.

This is a DRAFT pull request to gain feedback as to whether such an approach would fit into this library.

### DIFF
--- a/src/json-schema.hpp
+++ b/src/json-schema.hpp
@@ -119,12 +119,22 @@ extern json draft7_schema_builtin;
 typedef std::function<void(const json_uri & /*id*/, json & /*value*/)> schema_loader;
 typedef std::function<void(const std::string & /*format*/, const std::string & /*value*/)> format_checker;
 
+// Validation error information
+struct JSON_SCHEMA_VALIDATOR_API error_info
+{
+	json::json_pointer ptr;
+	std::string message;
+	std::vector<error_info> subschema_errors;
+	// formatted error information including ptr, message and subschema errors
+	std::string formatted() const;
+};
+
 // Interface for validation error handlers
 class JSON_SCHEMA_VALIDATOR_API error_handler
 {
 public:
 	virtual ~error_handler() {}
-	virtual void error(const json::json_pointer & /*ptr*/, const json & /*instance*/, const std::string & /*message*/) = 0;
+	virtual void error(const error_info & /*err*/) = 0;
 };
 
 class JSON_SCHEMA_VALIDATOR_API basic_error_handler : public error_handler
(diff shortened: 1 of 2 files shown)


garethsb-sony:
I suspect it would be a good idea to align `error_info` with the http://json-schema.org/work-in-progress#output-schema included in JSON-Schema draft08.

garethsb-sony:
I think the draft08 `"absoluteKeywordLocation"` can be generated from the schema locations and JSON pointers held in the `root_schema`. However, in the current implementation, during validation I think a `schema` cannot get these (they are stored in `root_schema` as map keys. I propose to hold these inside the `schema` objects as well as, or instead of (and use `set` with custom Comparator for example), in the `root_schema`.

On the other hand, the `"keywordLocation"` needs to include `$ref` explicitly based on how the validation is driven, which I think is best accomplished by adding a schema `json_pointer` to the `validate` function arguments.

garethsb-sony:
Another question to answer before I really get stuck into this is whether all the different 'draft08' validation output levels (`flag`, `basic`, `detail`, `verbose`)  make sense for this implementation and whether the gathering of annotations when the document is valid is also useful or not?

pboettch:
You are asking good questions here. I haven't looked at draf8 at all. I can't be of much help in this discussion for the moment.

garethsb-sony:
Thanks for the positive response, @pboettch. I've been diving into the definition of Output Formatting in the latest JSON Schema draft, and it makes a lot of sense.

I think we can end up with an implementation that is able to produce the final validation output in any of the four forms. There will need to be some structural changes to the code that will ensure the output includes all failures in the case of multiple failures *and* on the other hand, allow "short circuit" validation for performance in all cases where that's possible. Short circuiting has two forms:
* Skipping subsequent `"anyOf"` items after one succeeds
* Skipping subsequent `"allOf"` items (and keyword siblings) after one fails
The second form of short circuiting should only be performed for _Flag_ output.

> On the other hand, the `"keywordLocation"` needs to include `$ref` explicitly based on how the validation is driven, which I think is best accomplished by adding a schema `json_pointer` to the validate function arguments.

I note that the statement in https://github.com/json-schema-org/json-schema-spec/issues/779#issuecomment-520191039 reflects how I was thinking of tweaking the implementation of `schema_ref` to achieve the above!

(FWIW, I think most of the above will bring much better error reporting without  the need to adopt the other changes proposed in draft08 yet.)

lkersting:
Is there still work to include something like this?

pboettch:
I haven't checked the newer drafts/standards but back then I saw that error-reporting was defined as well. To make it clearer independently of the chosen programming language.

garethsb:
I haven't had time to investigate the progress of the spec or bring this to a conclusion, sorry. It's still a good idea! 💡

sam20908:
Is there any updates on this? This would be appreciated

PR #69, opened 2019-10-11
Fix non-compilable code example in README.md

44DD:
Code example in README.md does not compile.
First argument of basic_error_handler::error is json_pointer instead of std::string.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,7 @@ See also `app/json-schema-validate.cpp`.
 
 ```C++
 #include <iostream>
+#include <iomanip>
 
 #include "json-schema.hpp"
 
@@ -177,10 +178,11 @@ int main()
 	/* json-parse the people - with custom error handler */
 	class custom_error_handler : public nlohmann::json_schema::basic_error_handler
 	{
-		void error(const std::string &path, const json &instance, const std::string &message) override
+		void error(const nlohmann::json_pointer<nlohmann::basic_json<>> &pointer, const json &instance,
+			const std::string &message) override
 		{
-			nlohmann::json_schema::basic_error_handler::error(path, instance, message);
-			std::cerr << "ERROR: '" << path << "' - '" << instance << "': " << message << "\n";
+			nlohmann::json_schema::basic_error_handler::error(pointer, instance, message);
+			std::cerr << "ERROR: '" << pointer << "' - '" << instance << "': " << message << "\n";
 		}
 	};
 


README.md
```diff
@@ -112,6 +112,7 @@ See also `app/json-schema-validate.cpp`.
 
 ```C++
 #include <iostream>
+#include <iomanip>
```
pboettch: iomanip is needed? Where?

pboettch:
Also, you should use the same email address on github as the one you used for your commit. This way, github will associated your account with this contribution.

README.md
44DD: setw is defined in iomanip https://en.cppreference.com/w/cpp/io/manip/setw
`+--> g++ -c -o main.o main.cpp -I ../json-schema-validator/src/ -I ../json/include/ -std=c++1z
main.cpp: In function ‘int main()’:
main.cpp:55:21: error: ‘setw’ is not a member of ‘std’
             << std::setw(2) << person << std::endl;`

pboettch: Oops; yes, you're right, I thought it was an addition you made to the `std::cerr` statement. Sorry.

pboettch:
Great work. Depending on the project push-forcing is sometimes not wanted. Most of the time it is.

Github is well aware of push-forcing for updating a PR and handles comments on previous versions of the PR correctly. Which is nice.

So, cutting a long story short, push forcing is the right way for me.

Your commit seems now to contain two author-fields. Is this wanted?

Another way to allow Github associating commits to your github-account is to register multiple email-addresses to your account. Then you can push from work, from home, from Mars - github will accord changes to your account.

44DD:
> Great work. Depending on the project push-forcing is sometimes not wanted. Most of the time it is.
So usually I just make on-top commit and someone will squash when merging to master?

> Then you can push from work, from home, from Mars - github will accord changes to your account.
You got exact my case - I pushed from my work PC at Mars.

> Your commit seems now to contain two author-fields. Is this wanted?
TBH, I don't know how to change it. I'd leave it as is, if it's not a problem.

Thanks for github mentoring ;)

pboettch:
> > Great work. Depending on the project push-forcing is sometimes not wanted. Most of the time it is.
> 
> So usually I just make on-top commit and someone will squash when merging to master?

Both is fine actually. If incremental commits are just PR-review-fixes, they should be squashed when merged. Big developments, consisting of multiple useful commits should be merged, rebased as is. 

This is how I do.

> > Then you can push from work, from home, from Mars - github will accord changes to your account.
> 
> You got exact my case - I pushed from my work PC at Mars.

Say hi to Curiosity.

PR #71, opened 2019-10-17
Added Config file and updated install scripts.

lkersting:
Now when json-schema-validator is installed CMake config files are installed in the lib/cmake/json-schema-validator directory.
The install json-schema-validatorTargets.cmake file properly imports the json-hpp and json-schema-validator libraries.
The install json-schema-validatorConfig.cmake file is used by CMake find_package function to include the json-schema-validatorTargets.cmake file and to set the variable JSON_SCHEMA_VALIDATOR_INCLUDE_DIRS to the install include directory.
To use find_package to find the json-schema-validator simply include

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,7 +30,12 @@ if(NOT TARGET json-hpp)
     add_library(json-hpp INTERFACE)
     target_include_directories(json-hpp
         INTERFACE
-            ${NLOHMANN_JSON_REALPATH})
+            $<INSTALL_INTERFACE:include>
+            $<BUILD_INTERFACE:${NLOHMANN_JSON_REALPATH}>)
+
+    install(TARGETS json-hpp EXPORT ${PROJECT_NAME}Targets)
+    install(FILES ${NLOHMANN_JSON_REALPATH}/nlohmann/json.hpp DESTINATION include/nlohmann)
+
 endif()
 
 # and one for the validator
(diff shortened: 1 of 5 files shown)


lkersting:
This is basically the same as pull request #24 but this implementation uses more modern CMake. I am using CMake 3.12 and I haven't tested it with CMake 3.2.

pboettch:
I'm having a hardtime with these changes (and those from #24).

Could you write a test for it? Installing locally inside the build-dir and building an application with the installed lib/includes?

lkersting:
What kind of test are we talking about? I could write a bash test script that configures and builds json-schema-validator inside a build-dir and then installs it inside a installs-dir and then compiles a simple project that uses find_package to import the installed lib/includes. I'm not sure how to make that something ctest would run though.

pboettch:
Yes, a bash-script doing exactly what you describe would great :+1: This script would be `set -e` and manually integrated in ctest.

lkersting:
I've updated the merge request with the new test. It is passing for me, let me know if it works for you. The test may take a while since it has to configure, build, and install the validator. The new test test_cmake_install is a simple cmake project that runs the issue 9, 12, 27, 48, and 54 tests.

pboettch:
Thanks again for your work. Highly appreciated.

Based on your work, I took the occasion to change quite some stuff regarding finding nlohmann-json and some name-adaptations.

I pushed this on PR #73.

Could you please take a look? It contains your changes and my modifications.

Issue #79, opened 2020-01-18
CMake Error because of install(EXPORT ...) maybe turn into an option?


merlinmarek:
Hey there.
I have both, the validator and nlohmann_json as git submodules in my project. I build a binary that does not export nlohmann_json and the validator. When I want to generate the build files with CMAKE, I get the following error

`
-- Using the single-header code from /tmp/tmp.ksoqdZpHuy/nlohmann_json/single_include/
-- Found nlohmann_json::nlohmann_json-target - linking with it
-- Configuring done
CMake Error: install(EXPORT "nlohmann_json_schema_validatorTargets" ...) includes target "nlohmann_json_schema_validator" which requires target "nlohmann_json" that is not in any export set.
-- Generating done
CMake Generate step failed.  Build files cannot be regenerated correctly.
`

If I comment out the lines
`
install(EXPORT ${PROJECT_NAME}Targets
        FILE ${PROJECT_NAME}Targets.cmake
        DESTINATION "${INSTALL_CMAKE_DIR}")
`
at the very bottom of the CMakeLists.txt, I do not get the error and the build file generation succeeds. I am not a CMake expert, so maybe there is a better way to get around this.

I have prepared a dummy project that is setup the same way as my current real project [here](https://github.com/merlinmarek/json_schema_validator_issue).

How do you think about adding an option for the export (and maybe related functionality)?

pboettch:
Hmm, I'll take a look soon, I'm in the same situation for a project I haven't updated yet.

pboettch:
I found a solution, which is deactivating entirely the install-commands when being used as a submodule.

All these different ways of how to build the library is making harder and harder. Letting some time goes by I'm sure reducing the possibilities and forcing users into 2 or max 3 possibilities.

Instead of using add_subdirectory(), which pollutes a little bit your project, in the past I chose to use ExternalProject for including the validator lib.

merlinmarek:
#81 is exactly what I was looking for. Will this be merged?
I tried to setup the dependency using ExternalProjectAdd, but I am running into issues. I would prefer the add_subdirectory way, but out of curiosity, could you provide a minimal ExternalProjectAdd example?
I am using this:
`
ExternalProject_Add(json_schema_validator
	SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/json_schema_validator"
	CMAKE_CACHE_ARGS
	-DBUILD_TESTS:BOOL=OFF
[...]

PR #81, opened 2020-01-27
fix / workaround for #79

pboettch:


pboettch:
Could you check whether this works for you? @merlinmarek

PR #86, opened 2020-02-10
Avoid copy of json on set_root_schema

Finkman:
If a json object containing a schema is moved into the json validator it
should not be copied. At least the public interface of the validator
class schould not force the copy.

### DIFF
--- a/src/json-validator.cpp
+++ b/src/json-validator.cpp
@@ -1150,6 +1150,11 @@ void json_validator::set_root_schema(const json &schema)
 	root_->set_root_schema(schema);
 }
 
+void json_validator::set_root_schema(json &&schema)
+{
+	root_->set_root_schema(std::move(schema));
+}
+
 void json_validator::validate(const json &instance) const
 {
 	throwing_error_handler err;
--- a/src/nlohmann/json-schema.hpp
+++ b/src/nlohmann/json-schema.hpp
@@ -176,6 +176,7 @@ public:
 
 	// insert and set the root-schema
 	void set_root_schema(const json &);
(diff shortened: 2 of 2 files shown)


Finkman:
Sorry, about the first commit in the PR. I was struggling to sync my fork with your repo.
[These are my first steps on contributing at github]

pboettch:
juste rebase and push force to your branch. Github will handle it.

pboettch:
When the user is setting the root-schema (as JSON instance) it is copied because during processing it is modified.

As you can see, `set_root_schema` of the validator class takes a `const &` and then it this copied to the `root_schema`-class via its `set_root_schema` (which itself passes it around as `&`)

In what way does the explicit move help here?

Finkman:
Think about
```c++
json_validator validator{};
validator.set_root_schema(R"(
{
    "title": "A person",
    "properties": {
        "name": {
            "description": "Name",
            "type": "string"
        },
        "age": {
            "description": "Age of the person",
            "type": "number",
            "minimum": 2,
            "maximum": 200
        }
    },
    "required": [
                 "name",
                 "age"
                 ],
    "type": "object"
}

)"_json);
```
I came across this when I was loading a schema from file and only need it with the `validator`. So I made no sense to me to make a implicit copy of schema, but maybe I am wrong and it is not a use case at all.

In case `json` is *cheap-to-move*; another solution would be to remove `const&` and apply `pass by value and move`
https://stackoverflow.com/a/21035511/9233740

pboettch:
It's not _cheap-to-move_. Understood the use-case. Thanks for the example.

PR #89, opened 2020-03-03
Default generation

### LINKED ISSUE #25: Optionally add default values to json being validated when the field is missing
It would be really neat if, in the process of validation, the validator could (optionally) fill in missing values with their defaults if provided in the schema. The logic might look something like this:

```
IF element missing
    IF required THEN throw error
    ELSE IF default value exists THEN
        ADD the element with the specified default value
    ELSE skip to next element
END IF

VALIDATE element
```

This logic validates the element when it is entered or when it's replaced by a default value (unless it's either required or there is no default) which lets you validate your schema default values as well.

(Alternatively, default value validation could take place when the schema is added to the validator object, which is probably a good idea, but more work than using the above logic as a stop-gap that lets library users validate their default values by writing a test that simply does not supply them and expects the validation to pass.)

Finkman:
This is my contribution that targets #25. It brings almost no impact on the public interface which means that it would not break previous usage.

It adds a return value to the validation call which contains a patch (RFC 6902) which is supported by the nlohmann::json. See: [JSON Pointer and JSON Patch](https://github.com/nlohmann/json#json-pointer-and-json-patch)

I would appreciate comments, changes and suggestions a lot.

(Close #25)


CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ add_library(nlohmann_json_schema_validator
     src/json-schema-draft7.json.cpp
     src/json-uri.cpp
     src/json-validator.cpp
+    src/json_patch.cpp
```
pboettch: Please use - for filename conjunctions.

app/readme.cpp
```diff
@@ -21,7 +21,16 @@ static json person_schema = R"(
...
+        },
+				"address":{
+					"type": "object",
+					"properties":{
+						"street":{
+							"type": "string",
+							"default": "Abbey Road"
+						}
+					}
+				}
```
pboettch: No tabs in JSON-strings.

src/json-validator.cpp
```diff
@@ -783,38 +829,49 @@ class object : public schema
...
+		for (auto const &prop : properties_) {
+			const auto finding = instance.find(prop.first);
+			if (instance.end() == finding) { // if the prop is not in the instance
+				const auto &defaultValue = prop.second->defaultValue(ptr, instance, e);
+				if (!defaultValue.empty()) { // if default value is available
+					patch.add((ptr / prop.first), defaultValue);
+				}
+			}
+		}
+
```
pboettch: At first, years ago when I started the validator, I was wondering wether it is a not too heavy to carry around the json_pointer and updating it. Now I see that this helped to code your feature. Great.

CMakeLists.txt
Finkman: sorry. I tried to apply your conventions. I was not focused enough and saw an underscore.

src/json-validator.cpp
Finkman: Yes, that helped a lot.

Finkman:
> We would need some unit-tests ideally before merging.

Yes, of course. But I need a little time to investigate into cmake-tests.

Finkman:
Are more UnitTests are needed? What kind?

pboettch:
I approved the changes and then did a ctest: 

```
99% tests passed, 1 tests failed out of 70

Total Test time (real) =  12.35 sec

The following tests FAILED:
	  5 - issue-25-default-values (Failed)
Errors while running CTest
```

pboettch:
The is the JSON-patch I'm getting:
```JSON
[[],{"op":"add","path":"/address/street","value":"Abbey Road"}]
```

Finkman:
hmm.. that's odd.

```
...
[ctest] 67/69 Test  #5: issue-25-default-values ...............................   Passed    0.01 sec
...
```
for commit 6709194

pboettch:
What version of nlohmann::json? and what gcc/clang?

pboettch:
It works for me when I change the constructor of json_patch to ` j_(R"([])"_json)` instead of using the braces-initializer.

It also works, if I do not set an initializer at all.

Finkman:
> What version of nlohmann::json? and what gcc/clang?

I'm using json 3.7.3 and clang 8.0.1

Finkman:
Now, it should work-out fine. I am using the default initializer.

pboettch:
Integrated. Thanks a lot.

PR #95, opened 2020-03-26
adding CMAKE_INSTALL_PREFIX to install dirs

aschoenenberg:
adding CMAKE_INSTALL_PREFIX for cross-compiling


aschoenenberg:
You're right, was a problem on my machine

PR #107, opened 2020-04-19
using maybe_unused attribute only when supported

### LINKED ISSUE #106: Usage of [[maybe_unused]] limits compatibility
https://github.com/pboettch/json-schema-validator/blob/b28c15adea57bc9556cb00a7611536b5899dc67c/src/json-patch.cpp#L62

On visual studio 15 it get the following error...
```
C:\.conan\fc4f27\1\source_subfolder\src\json-patch.cpp(62): error C2429: attribute 'maybe_unused' requires compiler flag '/std:c++17' [C:\.conan\fc4f27\1\build_subfolder\source_subfolder\nlohmann_json_schema_validator.vcxproj]
```

prince-chrismc:
Making sure the current configuration supports [[maybe_unused]] before trying to compile with it

fixes #106 

I am not sure if there are any other usages of attributes this is the one that prevent me from  compiling

[build passing](https://travis-ci.com/github/prince-chrismc/json-schema-validator/builds/160911408)

### DIFF
--- a/src/json-patch.cpp
+++ b/src/json-patch.cpp
@@ -1,5 +1,15 @@
 #include "json-patch.hpp"
 
+#ifdef __has_cpp_attribute
+#if __has_cpp_attribute(maybe_unused)
+#define JSV_MAYBE_UNUSED [[maybe_unused]]
+#endif
+#endif
+
+#ifndef JSV_MAYBE_UNUSED
+#define JSV_MAYBE_UNUSED
+#endif
+
 namespace nlohmann
 {
 
@@ -59,7 +69,7 @@ void json_patch::validateJsonPatch(json const &patch)
 
 		try {
 			// try parse to path
-			[[maybe_unused]] const auto p = json::json_pointer{op["path"].get<std::string>()};
+			JSV_MAYBE_UNUSED const auto p = json::json_pointer{op["path"].get<std::string>()};
 		} catch (json::exception &e) {
 			throw JsonPatchFormatException{e.what()};
 		}
--- a/src/json-patch.cpp
+++ b/src/json-patch.cpp
@@ -1,15 +1,5 @@
 #include "json-patch.hpp"
 
-#ifdef __has_cpp_attribute
-#if __has_cpp_attribute(maybe_unused)
-#define JSV_MAYBE_UNUSED [[maybe_unused]]
-#endif
-#endif
-
-#ifndef JSV_MAYBE_UNUSED
-#define JSV_MAYBE_UNUSED
-#endif
-
 namespace nlohmann
 {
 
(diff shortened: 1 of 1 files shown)


src/json-patch.cpp
```diff
@@ -59,7 +69,7 @@ void json_patch::validateJsonPatch(json const &patch)
 
 		try {
 			// try parse to path
-			[[maybe_unused]] const auto p = json::json_pointer{op["path"].get<std::string>()};
+			JSV_MAYBE_UNUSED const auto p = json::json_pointer{op["path"].get<std::string>()};
```
pboettch, line 72: What about using the old-fashioned `(void)`-cast to eliminate this warning?

prince-chrismc: ~~I think it's only a warning when the compiler supports this attribute =?~~

Playing with this on [compiler explorer](https://godbolt.org/z/iYJywL) it seems compilers with less then C++17 STL support (which this feature is apart of) compile but still produce a warning regardless.

prince-chrismc:
```suggestion
			const auto p = json::json_pointer{op["path"].get<std::string>()}; p;
```

pboettch: And without using a variable at all?

prince-chrismc: .... brillant !

prince-chrismc:
```suggestion
			json::json_pointer{op["path"].get<std::string>()};
```

pboettch: Yeah, but does it still do the job - ie. doesn't the compiler throw away the code because of not using the instance.

prince-chrismc: Sadly yes =?

pboettch:
So, does this actually work?

prince-chrismc:
Sadly no. 

The best options
1. return the value (no requirement to check) 
2. use the attribute where it's available (compilers without the support with produce a warning) but since this is a compile `cpp` consumers like myself won't really care.

Otherwise there's a risk of it being optimized away.

pboettch:
Strange, the compiler is not allow to remove a call even to a constructor. He cannot now whether the code in the constructor does something on global scale or not.

pboettch:
Lesson learnt : when there is no unit-test it doesn't work and it can't be verifed.

I rewrite the validation path of the json-patch using a json-schema!

And I added a unit-test which validated some cases. 

I we see that 

```C++
	for (auto const &op : patch)
		json::json_pointer(op["path"].get<std::string>());
```
works.

No need to for `maybe_unused`.

prince-chrismc:
> when there is no unit-test it doesn't work and it can't be verifed.

Lesson I have been learning my whole career :exploding_head: 
As I use you library more, I hope my PRs will be more helpful


> Strange, the compiler is not allow to remove a call even to a constructor. He cannot now whether the code in the constructor does something on global scale or not.

My suspicious was the _dead code elimination_ could tag this code since its only writing to a variable that's never used. Typically this is seen when reducing for size.

pboettch:
Compiler can't do it. `json::json_pointer(op["path"].get<std::string>())`-code is in another library (or could) and thus cannot be inlined.

A function call is never eliminated as dead-code, except if inlined or complete visibility is given.

PR #115, opened 2020-05-25
Add functions to get library version at runtime

xamix:
This add functions to get current library version.
Allow to retrieve MAJOR, MINOR, PATCH and a concatenated version string

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,12 +1,17 @@
-project(nlohmann_json_schema_validator
-        LANGUAGES CXX)
-
-set(PROJECT_VERSION 2.1.1)
 
 cmake_minimum_required(VERSION 3.2)
+cmake_policy(SET CMP0048 NEW) # Needed to set version with project
+
+project(nlohmann_json_schema_validator
+        VERSION 2.1.1
+        LANGUAGES CXX)
 
 option(BUILD_TESTS    "Build tests"    ON)
 option(BUILD_EXAMPLES "Build examples" ON)
+
+# generate file with version
+configure_file ("${CMAKE_CURRENT_SOURCE_DIR}/src/nlohmann/json-schema.hpp.in"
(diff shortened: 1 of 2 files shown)


xamix:
Hi, 
In fact I think it will be usefull to get it also at compile time, since I will not compile my project if version is lesser than the new 2.1.1.

To do that I need to modify the `json-schema.hpp` only.

Maybe can I add only some define in the current modifyed `json-schema.hpp` file instead of funtion and use thoses defines also at runtime:

```
#define SCHEMA_VALIDATOR_VERSION_MAJOR  2
#define SCHEMA_VALIDATOR_VERSION_MINOR  1
#define SCHEMA_VALIDATOR_VERSION_PATCH  1
#define SCHEMA_VALIDATOR_VERSION_STRING "2.1.0"
```

Finkman:
> In fact I think it will be usefull to get it also at compile time

In our projects, we use a `Version.hpp` which is generated by CMake which includes the correct version information on each configure run. With that you can put the version information to header without hard-coding it.
But: To check if a lib is suitable, you should check for the correct version in the cmake configuration.

Moreover, there is always a chance, that the header does not match the version of the binary. But for that, there is a solution as well. The pedantic solution is to implement a template specialization for a check function:

```cpp
namespace lib_namespace {
  static const auto version_ok __attribute__( ( unused ) ) =
     lib_namespace::about::check_library_version<  lib_namespace::VERSION_MAJOR,  lib_namespace::VERSION_MINOR >();
}
```
where `VERSION_MAJOR` and `VERSION_MINOR` is the generated version info mentioned above.
The binary of the lib then implements the above check for only the correct version numbers. If there is a mismatch between header and binary, it results in a linker error.

If @pboettch is interessted in that, I can contribute.

pboettch:
I like your idea @Finkman. Do not hesitate to make a PR.

Finkman:
Take a look at #139

pboettch:
Closing this one, as #139 will make it into the library soon.

PR #117, opened 2020-06-04
add checking about format-checker existance at schema parsing time

andrejlevkovitch:
Now if you use schema with `format` field and don't set format checker, then you get error about not provided `format checker` at validation time. But what about verify existence of `format checker` at schema parsing time? I think it will be more useful

### DIFF
--- a/src/json-validator.cpp
+++ b/src/json-validator.cpp
@@ -672,6 +672,10 @@ public:
 
 		attr = sch.find("format");
 		if (attr != sch.end()) {
+			if (root_->format_check() == nullptr) {
+				throw std::invalid_argument{"a format checker was not provided but a format keyword for this string is present: " + format_.second};
+			}
+
 			format_ = {true, attr.value()};
 			sch.erase(attr);
 		}
--- a/src/json-validator.cpp
+++ b/src/json-validator.cpp
@@ -672,9 +672,8 @@ public:
 
 		attr = sch.find("format");
 		if (attr != sch.end()) {
-			if (root_->format_check() == nullptr) {
+			if (root_->format_check() == nullptr)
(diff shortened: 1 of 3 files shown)


test/issue-117-format-error.cpp
```diff
@@ -0,0 +1,69 @@
...
+			x;                                                        \
+		} catch (std::exception & e) {                                \
+			fail = true;                                              \
+			exception_error = e.what();                               \
+		}                                                             \
+		if (fail == true) {                                           \
+			++error_count;                                            \
+			std::cout << msg << ": " << exception_error << std::endl; \
+		}                                                             \
+	}
```
pboettch, line 37: We should share this kind of defines between all the tests.

src/json-validator.cpp
```diff
@@ -672,6 +672,10 @@ class string : public schema
 
 		attr = sch.find("format");
 		if (attr != sch.end()) {
+			if (root_->format_check() == nullptr) {
+				throw std::invalid_argument{"a format checker was not provided but a format keyword for this string is present: " + format_.second};
+			}
+
```
pboettch, line 678: Without `{}` please. Single-line-if.

test/issue-117-format-error.cpp
andrejlevkovitch: sorry, I didn't understand you

pboettch: I mean, these macros could go into a seperate .h-file and be used by others tests.

andrejlevkovitch: Yes, of course, but maybe better use some complete test library? For example Catch2

pboettch: Yes, you're right, I never used it, but it seems to be time to use it here.

PR #119, opened 2020-06-10
Disable tests and examples when used via add_subdirectory

jakobandersen:
Similarly to how installation is disabled when included via ``add_subdirectory``, I believe test and example building should be disabled as well.

It seems that CMake does not make it possible to set options only for a sub-project without affecting the parent project. An alternative to this PR would be to rename ``BUILD_TESTS`` and ``BUILD_EXAMPLES`` to be globally unique so they can be set in the parent.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,6 +45,8 @@ if(TARGET nlohmann_json::nlohmann_json)
         PUBLIC nlohmann_json::nlohmann_json)
 
     set(JSON_VALIDATOR_INSTALL OFF)
+    set(BUILD_TESTS OFF)
+    set(BUILD_EXAMPLES OFF)
 
 elseif(TARGET nlohmann_json) # or nlohmann_json, we are used a sub-project next to nlohmann-json's git repo
     message(STATUS "Found nlohmann_json-target - linking with it")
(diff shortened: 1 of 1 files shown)


pboettch:
In one project of mine I actually used the tests of the validation in a parent project.

It might be a rare case where this is useful. 

Regarding setting BUILD_TESTS before `add_subdirectory` does this not work ? (It should)
```CMake
set(BUILD_TESTS OFF) # maybe with CACHE
add_subdirectory(...)
set(BUILD_TESTS ON)
```

jakobandersen:
As you hint, the ``CACHE`` part of ``SET`` seems to be needed, as ``option`` adds the variable as a cached variable. The problem is that I use the same name for options in my parent project, so I would disable my own tests/examples as well.
Without the ``CACHE``  part it seems to work on subsequent cmake invocations, but the first clean one seems to fails in some way.

How about leaving the options as is (I don't like the thought of option names being required to be unique across unrelated projects) and then still disabling examples/test as default when added as a subdirectory, but then have two new variables with more obscure names to forcibly reenable them anyway?
From the perspective of a parent project the behaviour is as default the same whether the import is via ``add_subdirectory`` or ``find_package``.

pboettch:
No, you're right, disabling it while using the validator a sub-dir is the right way. The rare case I mentioned is not enough to justify an even more complicated CMakeLists.txt

PR #123, opened 2020-06-22
fix default is empty object

gzzi:
Hello,

This fix resolve a issue we faced with empty object as default value. Previous behavior of null as default value is the same (is_empty on a null json return true).

Here is a example of new behavior:

Schema:
```
{
    "title": "A person",
    "properties": {
      "address":{
        "type": "object",
        "default": {}
      }
    },
    "type": "object"
}
```

Object:
`{}`

Patch:
```
[{"op":"add","path":"/address","value":{}}]
```

Object after applying patch:
```
{
  "address": {}
}
```

### DIFF
--- a/test/issue-25-default-values.cpp
+++ b/test/issue-25-default-values.cpp
@@ -19,14 +19,15 @@ static const json person_schema = R"(
             "minimum": 2,
             "maximum": 200
         },
-            "address":{
-                "type": "object",
-                "properties":{
-                    "street":{
-                        "type": "string",
-                        "default": "Abbey Road"
+        "address": {
+            "type": "object",
+            "default": {},
+            "properties": {
+                "street": {
+                    "type": "string",
+                    "default": "Abbey Road"
+                }
             }
-          }
         }
     },
     "required": [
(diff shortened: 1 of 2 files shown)


gzzi:
agree, see new commit

test/issue-25-default-values.cpp
```diff
@@ -28,6 +28,25 @@ static const json person_schema = R"(
...
+        "work address": {
+            "type": "object",
+            "default": null,
+            "properties": {
+                "street": {
+                    "type": "string",
+                    "default": "Abbey Road"
+                }
+            }
+        },
```
pboettch, line 41: This is interesting! The parent object of `street` has default = null, but street has a non-null default.

Why do we assume that a parent default is more important than a child one? Same as below: no parent-default means child defaults are ignored.

This is an issue which needs to be looked into.

pboettch:
@Finkman could you please review this?

test/issue-25-default-values.cpp
pboettch: https://github.com/pboettch/json-schema-validator/issues/124

gzzi: Not sure it is a issue. Let me explain a bit our context:

My journey started when I was willing to generate default out of schema (I have seen, even if it is pure annotation, that need seem common). I looked to this library we were already using and discovered that it generate a patch with default value as long as the parent object exist. So to generate default, I do:
 - remove all required inside the schema
 - start with empty object
 - validate with schema
 - apply patch
 - repeat until patch is a empty list

And it work fine. It is a bit strange and a method on the library that do this job would be great but at the end it work.

Of course, another way would be to repeat all member on the default object up to root. But repeating over and over values doesn't sound good to me.

Finkman:
I fully agree with the changes.
In my point of view, the default value of a parent is superior to the children. The reason for it is that the children my come from a referenced definition and my be used in more than one contexts which might have different requirements on default values.

But, my option might not be the reference for it. There are some guys around which are more responsible for that kind of philosophy.

PR #121, opened 2020-06-15
Hunter package manager support and some fixes 

themarpe:
Hi.
This PR adds a couple of changes.

- Hunter package manager. It is dependent only on CMake. Right now it's disabled by default, but can be enabled using `-DHUNTER_ENABLED=ON`.
Using Hunter, there is no need to have Nlohmann JSON target available.

- Fix which was needed to make install work under Hunter and I think also fixes issue #77 

- MSVC compilation error fix (src/json-validator.cpp)

I'm also interested in adding this package to Hunter - this will enable users to use this library with ease

PR #126, opened 2020-07-19
Fixed Hunter integration and added tests

themarpe:
This is a follow up to PR #121. It fixes the integration, adds travis tests using Hunter package manager to retrieve nlohmann json library. 
It also includes the MSVC fix.
Branch was rebased to master and commits squashed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,3 +67,12 @@ script:
   - cmake .. -Dnlohmann_json_DIR=.. ${CMAKE_OPTIONS} -GNinja && cmake --build . --config Release
   - ctest -C Release -V -j
   - cd ..
+
+  # Remove previous build and tests
+  - rm -r build
+
+  # Compile and execute with Hunter package manager instead of using local json.hpp
+  - mkdir -p build && cd build
+  - cmake .. -DHUNTER_ENABLED=ON ${CMAKE_OPTIONS} -GNinja && cmake --build . --config Release
+  - ctest -C Release -V -j
+  - cd ..
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,3 +1,13 @@
+# Add Hunter support (Disabled by default)
+option(HUNTER_ENABLED "Enable Hunter package manager support" OFF)
+if(HUNTER_ENABLED)
+    include("cmake/HunterGate.cmake")
+    HunterGate(
+        URL "https://github.com/cpp-pm/hunter/archive/v0.23.262.tar.gz"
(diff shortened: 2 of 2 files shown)


src/json-validator.cpp
```diff
@@ -727,7 +727,7 @@ class string : public schema
 #ifndef NO_STD_REGEX
 		attr = sch.find("pattern");
 		if (attr != sch.end()) {
-			patternString_ = attr.value();
+			patternString_ = attr.value().get<std::string>();
```
pboettch, line 730: What is the error message you're receiving for this `get<>()`-call? Which version of the compiler are you using?

I'm asking because the compiler should determine automatically the requested type.

themarpe: I tried to recreate the issue, but it seems that using v3.8.0 nlohmann json.hpp and latest commit it works.
I'll remove this and check into adding MSVC & Clang compilers into CI to minimize this kind of errors in future.

For reference: 
Using v3.7.3 json.hpp and commit (a0fca479f60235ed775b7aac8c441688466c3a76) on Visual Studio 2019 it produces the following error: 
```
2> ...\json-schema-validator\src\json-validator.cpp(660,1): error C2593: 'operator =' is ambiguous
2>C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.25.28610\include\xstring(2591,19): message : could be 'std::basic_string<char,std::char_traits<char>,std::allocator<char>> &std::basic_string<char,std::char_traits<char>,std::allocator<char>>::operator =(std::initializer_list<_Elem>)'
2>        with
2>        [
2>            _Elem=char
2>        ]
2>C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.25.28610\include\xstring(2492,19): message : or       'std::basic_string<char,std::char_traits<char>,std::allocator<char>> &std::basic_string<char,std::char_traits<char>,std::allocator<char>>::operator =(std::basic_string<char,std::char_traits<char>,std::allocator<char>> &&) noexcept(<expr>)'
2> ...\json-schema-validator\src\json-validator.cpp(660,1): message : while trying to match the argument list '(std::string, nlohmann::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::adl_serializer>)'
2>json-patch.cpp
```

pboettch, line 730:
```suggestion
			patternString_ = attr.value();
```

themarpe:
Updated the PR with above discussed changes. Sorry for the delay.

PR #128, opened 2020-07-20
re-worked install configs

### LINKED ISSUE #127: Installed cmake config uses absolute paths
json-schema-validator uses absolute path to the installed files.
Instead it should use CMAKE_FIND_ROOT_PATH to allow to move installed files to a different location like it is done on cross compile toolchains.

Finkman:
Fixes #127 

- Add relative paths to cmake generated package configs
- Add nlohmann json 3.8. requirement to package config

Tested native and on cross-compile toolchain


Finkman:
> Does this change le minimum required CMake version? The usage of `CMakePackageConfigHelpers` for example.

You used it too: https://github.com/pboettch/json-schema-validator/pull/128/files#diff-af3b638bc2a3e6c650974192a53c7291L175

And I can see no other impacts on the required cmake version.

PR #130, opened 2020-09-02
Don't shadow "schema" with local variables of the same name

mm120:
This fixes a warning from gcc (-Wshadow).

We want to compile this project as part of a much larger internal project, and that uses -Wshadow.  We also have a zero-warnings policy, so it would be nice if these warnings could be silenced here.


pboettch:
Thanks. could you add the `-Wshadow` to the CMakeLists.txt (with `PRIVATE`) so that it is activated by default for this project.

pboettch:
I took the liberty to change some variable names. Thanks for your contribution.

mm120:
Many thanks!

PR #147, opened 2021-03-08
Prefer nlohamnn_json submodule

vaibhav-hexa:
Hello,
The changes in `CMakeLists` now allow to find `nlohmann_json` from the `submodule` instead of checking the path `/usr/local/include/nlohmann/`. This preference is required if an older version of `nlohmann_json` already exists and the `build` fails due to the minimum version check of `3.8.0` on nlohmann_json.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,3 +1,4 @@
+# Prefers json submodule
 # Add Hunter support (Disabled by default)
 option(HUNTER_ENABLED "Enable Hunter package manager support" OFF)
 if(HUNTER_ENABLED)
@@ -19,9 +20,9 @@ option(BUILD_TESTS    "Build tests"    ON)
 option(BUILD_EXAMPLES "Build examples" ON)
 
 # Add nlohmann_json::nlohmann_json using Hunter package manager
-# if(HUNTER_ENABLED)
-#     hunter_add_package(nlohmann_json)
-# endif()
+if(HUNTER_ENABLED)
+    hunter_add_package(nlohmann_json)
+endif()
 
 include_directories(../json/single_include)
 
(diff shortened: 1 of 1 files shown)


pboettch:
While this seems to fix your problem, the way you wrote your patch will break it for many other. What is the problem you're are trying to solve exactly.

Have you tried setting the `nlohmann_json_DIR`-cmake variable from the outside of your `add_subdirectory()`-call?

pboettch:
Could you show the code of your CMakeLists.txt which includes this project and you variables around this inclusion?

vaibhav-hexa:
> Could you show the code of your CMakeLists.txt which includes this project and you variables around this inclusion?

Please see the below code snippet.

```
...
# JSON SCHEMA VALIDATOR library
option(BUILD_TESTS OFF)
set(nlohmann_json_DIR "${CMAKE_CURRENT_SOURCE_DIR}/external/json/include")
add_subdirectory(external/json-schema-validator EXCLUDE_FROM_ALL)
include_directories(external/json-schema-validator/src)

# JSON library
option(JSON_BuildTests OFF)
add_subdirectory(external/json EXCLUDE_FROM_ALL)
include_directories(external/json/single_include)

..
..
..

# Install the header only json container
install(DIRECTORY external/json/single_include
  DESTINATION include
  FILES_MATCHING PATTERN "*.hpp"
)

# Install the header only json-schema container
install(DIRECTORY external/json-schema-validator/src
  DESTINATION include
  FILES_MATCHING PATTERN "*.hpp"
)
...
```

pboettch:
I created an example CMakeLists which is using the same two sub-directories as you are and it works.

```CMake
cmake_minimum_required(VERSION 3.2)

project(simple-json-validator-as-submodule-test)

# absolute-path
set(PATH_TO_NLOHMANN_JSON_REPO $ENV{HOME}/devel/upstream/json)

# build shared library
set(BUILD_SHARED_LIBS ON)

# JSON library
option(JSON_BuildTests OFF)
add_subdirectory(${PATH_TO_NLOHMANN_JSON_REPO}
                 json-binary-dir
                 EXCLUDE_FROM_ALL)

# JSON SCHEMA VALIDATOR library
option(BUILD_TESTS OFF)
add_subdirectory(../..
                 json-validator-binary-dir
                 EXCLUDE_FROM_ALL)


add_executable(validate validate.cpp)
target_link_libraries(validate nlohmann_json_schema_validator)
```

The key difference is: include nlohmann-json _before_ json-schema-validator.

Do _NOT_ use `include_directories()` in your project. It pollutes include-namespaces. Use `target_include_directories()`.

Could you please try it?

vaibhav-hexa:
Thank you for your suggestions, it works for me.

Issue #144, opened 2021-03-06
Unit tests fail on macOS big sur


Spuriosity1:
I recently tried to install this library by building from source.

System configuration:
macOS 11.2 Big Sur
nlohmann/json installed via Homebrew: `brew install nlohmann-json`, installed version 3.9.1
Output of `ls /usr/local/lib | grep json`:
`libnlohmann_json_schema_validator.a`
Output of `ls /usr/local/include/nlohmann`:
```
adl_serializer.hpp              detail                          json.hpp                        ordered_map.hpp
byte_container_with_subtype.hpp json-schema.hpp                 json_fwd.hpp                    thirdparty
```

Build sequence:
```
git clone https://github.com/pboettch/json-schema-validator.git
mkdir build && cd build
cmake ..
make
make install
ctest
```
giving output:
```
37% tests passed, 52 tests failed out of 82

Total Test time (real) =  48.86 sec

The following tests FAILED:
	 11 - JSON-Suite::additionalItems (Failed)
	 12 - JSON-Suite::additionalProperties (Failed)
	 13 - JSON-Suite::allOf (Failed)
	 14 - JSON-Suite::anyOf (Failed)
	 15 - JSON-Suite::boolean_schema (Failed)
	 16 - JSON-Suite::const (Failed)
	 17 - JSON-Suite::contains (Failed)
	 18 - JSON-Suite::default (Failed)
	 19 - JSON-Suite::definitions (Failed)
	 20 - JSON-Suite::dependencies (Failed)
	 21 - JSON-Suite::enum (Failed)
	 22 - JSON-Suite::exclusiveMaximum (Failed)
	 23 - JSON-Suite::exclusiveMinimum (Failed)
	 24 - JSON-Suite::format (Failed)
	 25 - JSON-Suite::if-then-else (Failed)
	 26 - JSON-Suite::items (Failed)
	 27 - JSON-Suite::maxItems (Failed)
	 28 - JSON-Suite::maxLength (Failed)
	 29 - JSON-Suite::maxProperties (Failed)
	 30 - JSON-Suite::maximum (Failed)
	 31 - JSON-Suite::minItems (Failed)
	 32 - JSON-Suite::minLength (Failed)
	 33 - JSON-Suite::minProperties (Failed)
	 34 - JSON-Suite::minimum (Failed)
	 35 - JSON-Suite::multipleOf (Failed)
	 36 - JSON-Suite::not (Failed)
[...]

Issue #141, opened 2021-01-23
Ctest fails on 51 tests after successful installation


grbba:
Hello,

I followed the installation using what i got from the readme file

```
git clone https://github.com/pboettch/json-schema-validator.git
cd json-schema-validator
mkdir build
cd build
cmake [..]
make
make install # if needed
ctest # run unit, non-regression and test-suite tests
```
everything completed successfully except the ctest once run reports 51 failed tests cf the attached file [ctest-fail.txt](https://github.com/pboettch/json-schema-validator/files/5861061/ctest-fail.txt)

I have a JSON Schema which seems to validate properly in VSC and the json files based upon this schema as well. I also have a program using the nlohmann parser and everything works reading & parsing the the json files. I wanted nevertheless add the validation in the program as users do not necessarily use a validating editor like VSC. 

loading/validating the schema actually works fine but when validating e json file then produces an : no subschema has succeeded, but one of them is required to validate error. 

Looking in the code i saw this has something to do with anyOf/oneOf and the anyOf test has failed ( as of the error file i added ) maybe both are linked?

I am on a Mac MacOS 10.15.7 using clang 
```
% clang --version
Apple clang version 12.0.0 (clang-1200.0.32.28)
Target: x86_64-apple-darwin19.6.0
Thread model: posix
```
I have to admit that i am a total beginner in this field and my schema / json files build upon that may actually be wrong somewhere but I still have to figure out how to debug those things properly.

Any help is appreciated very much.

pboettch:
You seem to have two different problems. One is that maybe your schema isn't correct so your validation fails.

The other one is that the non-regression tests fail. Let's focus first on the non-regs tests: could you run:

```
ctest -R JSON-Suite::type -V
```

and paste the output here?

grbba:
Hello,
thanks for your reply and yes there are  two different issues.
The result of the above command is :
```
UpdateCTestConfiguration  from :/Users/gregor/Development/json-schema-validator/build/DartConfiguration.tcl
UpdateCTestConfiguration  from :/Users/gregor/Development/json-schema-validator/build/DartConfiguration.tcl
Test project /Users/gregor/Development/json-schema-validator/build
Constructing a list of tests
Done constructing a list of tests
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
test 45
    Start 45: JSON-Suite::type

[...]

PR #150, opened 2021-03-09
Minor patch for bash 3.2 compatibility

Spuriosity1:
Default bash on macOS is bash 3.2, which does not support negative array indexing causing unit tests to appear to fail (#144 , #141). This syntax (copied from https://unix.stackexchange.com/a/198788/178794) corrects the issue by avoiding the use of a negative index.


pboettch:
I manually merged your changeset. Thanks a lot for this contribution.

PR #151, opened 2021-03-15
fixed occurences of old-style casts and an implicit cast to double

eike-fokken:
Hi!

We are using your validator but use more warnings than you do. In case you're interested, here I changed all casts (to uint8_t) to explicit static_casts and added a cast here:
```
-		double eps = std::nextafter(x, 0) - x;
+		double eps = std::nextafter(x, 0) - static_cast<double>(x);
```

In case you're not interested, please let me know.

By the way:
The validation library actually builds with the following warnings after these changes, although the tests do not (which may be due to nlohmann/json tests not building with these)
```
-Werror -Wconversion -pedantic-errors -pedantic -Wall -Wextra -Wcast-align -Wcast-qual -Wctor-dtor-privacy
 -Wdisabled-optimization -Wformat=2 -Winit-self -Wmissing-declarations -Wmissing-include-dirs -Wold-style-cast 
-Woverloaded-virtual -Wredundant-decls -Wshadow -Wsign-conversion -Wsign-promo -Wstrict-overflow=5
 -Wswitch-default -Wundef -Wunused -Wno-error=unused-parameter -Wno-error=unused-variable
 -Wno-error=unused-value -Wdelete-non-virtual-dtor
```

Thank you very much for your library!

### DIFF
--- a/src/json-uri.cpp
+++ b/src/json-uri.cpp
@@ -35,7 +35,7 @@ void json_uri::update(const std::string &uri)
 			}
 
 			std::string hex = pointer.substr(pos + 1, 2);
-			char ascii = static_cast<char>( std::strtoul(hex.c_str(), nullptr, 16));
+			char ascii = static_cast<char>(std::strtoul(hex.c_str(), nullptr, 16));
 			pointer.replace(pos, 3, 1, ascii);
 
 			pos--;
--- a/src/json-uri.cpp
+++ b/src/json-uri.cpp
@@ -35,7 +35,7 @@ void json_uri::update(const std::string &uri)
 			}
 
 			std::string hex = pointer.substr(pos + 1, 2);
-			char ascii = (char) std::strtoul(hex.c_str(), nullptr, 16);
+			char ascii = static_cast<char>( std::strtoul(hex.c_str(), nullptr, 16));
 			pointer.replace(pos, 3, 1, ascii);
 
 			pos--;
--- a/src/json-validator.cpp
+++ b/src/json-validator.cpp
@@ -456,7 +456,7 @@ class type_schema : public schema
 	void validate(const json::json_pointer &ptr, const json &instance, json_patch &patch, error_handler &e) const override final
 	{
 		// depending on the type of instance run the type specific validator - if present
-		auto type = type_[(uint8_t) instance.type()];
+		auto type = type_[static_cast<uint8_t>(instance.type())];
 
 		if (type)
 			type->validate(ptr, instance, patch, e);
@@ -500,7 +500,7 @@ public:
 	type_schema(json &sch,
 	            root_schema *root,
 	            const std::vector<nlohmann::json_uri> &uris)
-	    : schema(root), type_((uint8_t) json::value_t::discarded + 1)
+    : schema(root), type_(static_cast<uint8_t>(json::value_t::discarded) + 1)
 	{
 		// association between JSON-schema-type and NLohmann-types
 		static const std::vector<std::pair<std::string, json::value_t>> schema_types = {
@@ -518,7 +518,7 @@ public:
 		auto attr = sch.find("type");
 		if (attr == sch.end()) // no type field means all sub-types possible
 			for (auto &t : schema_types)
-				type_[(uint8_t) t.second] = type_schema::make(sch, t.second, root, uris, known_keywords);
+				type_[static_cast<uint8_t> (t.second)] = type_schema::make(sch, t.second, root, uris, known_keywords);
 		else {
 			switch (attr.value().type()) { // "type": "type"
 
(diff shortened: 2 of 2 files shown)


pboettch:
Harnessing the code is always appreciated.

Adding more warnings is a good idea, however, different compiler versions are moaning differently with some of them.

Maybe adding an option saying "more warnings" which can be enabled would be an acceptable alternative.

src/json-uri.cpp
```diff
@@ -35,7 +35,7 @@ void json_uri::update(const std::string &uri)
 			}
 
 			std::string hex = pointer.substr(pos + 1, 2);
-			char ascii = (char) std::strtoul(hex.c_str(), nullptr, 16);
+			char ascii = static_cast<char>( std::strtoul(hex.c_str(), nullptr, 16));
```
pboettch, line 38: The spaces after the `(` is too much. Fix all of them using clang-format if you can.

src/json-validator.cpp
```diff
@@ -518,22 +518,22 @@ class type_schema : public schema
 		auto attr = sch.find("type");
 		if (attr == sch.end()) // no type field means all sub-types possible
 			for (auto &t : schema_types)
-				type_[(uint8_t) t.second] = type_schema::make(sch, t.second, root, uris, known_keywords);
+				type_[static_cast<uint8_t> (t.second)] = type_schema::make(sch, t.second, root, uris, known_keywords);
```
pboettch, line 521: space as well.

pboettch (changes requested):

eike-fokken:
By the way the clang-format also moved the `#include <deque>`.

As for compiler problems, we use this warning setup, which works with clang, gcc and msvc (although the warnings on the last one are not really that finely tuned).
Here `my_debug_options` is a set of warnings that is understood by both clang and gcc and compiler-specific warnings are added afterwards.


```
set(my_debug_options -O0 -Werror -Wconversion -pedantic-errors -pedantic -Wall -Wextra -Wcast-align -Wcast-qual
-Wctor-dtor-privacy -Wdisabled-optimization -Wformat=2 -Winit-self -Wmissing-declarations -Wmissing-include-dirs 
-Wold-style-cast -Woverloaded-virtual -Wredundant-decls -Wshadow -Wsign-conversion -Wsign-promo 
-Wstrict-overflow=5 -Wswitch-default -Wundef -Wunused -Wno-error=unused-parameter -Wno-error=unused-variable 
-Wno-error=unused-value -Wdelete-non-virtual-dtor)

set(debug_clang_options "-fcolor-diagnostics;-Wno-error=unused-private-field;-Wmissing-prototypes")
set(debug_gcc_options "-fdiagnostics-color;-Wno-error=strict-overflow")

# /Zc:__cplusplus: https://gitlab.com/libeigen/eigen/-/issues/2170 might become superflous
set(debug_msvc_options "/permissive-;/W4;/WX;/wd4127;/Zc:__cplusplus")

set(release_clang_options "")
set(release_gcc_options "")
set(release_msvc_options "/permissive-")

if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
  add_compile_options("$<$<CONFIG:DEBUG>:${my_debug_options}>")
  add_compile_options("$<$<CONFIG:DEBUG>:${debug_clang_options}>"
    "$<$<NOT:$<CONFIG:Debug>>:${release_clang_options}>")
elseif (CMAKE_CXX_COMPILER_ID MATCHES "GNU")
  add_compile_options("$<$<CONFIG:DEBUG>:${my_debug_options}>")
   add_compile_options("$<$<CONFIG:DEBUG>:${debug_gcc_options}>"
     "$<$<NOT:$<CONFIG:Debug>>:${release_gcc_options}>")
elseif (CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
  add_compile_options(
    "$<$<CONFIG:DEBUG>:${debug_msvc_options}>"
    "$<$<NOT:$<CONFIG:Debug>>:${release_msvc_options}>"
  )
else()
  message(FATAL_ERROR "Unknown compiler: ${CMAKE_CXX_COMPILER_ID}")
endif()

```

pboettch (approved):

eike-fokken:
I have another remark for this:
Maybe (probably!) the change
```
double eps = std::nextafter(x, 0) - static_cast<double>(x);
```

was a bad one. I just realized that I got rid of the warning but didn't understand the code. So my question is: Have you considered, whether this cast is ok there?
What types T are actually instantiated there and will their cast to double ever become a problem (like when x is a really big long int, see here: [Stackoverflow](https://stackoverflow.com/questions/33824954/cast-from-unsigned-long-long-to-double-and-vice-versa-changes-the-value) )

So long story short: Please make really sure, my reckless doing away with the warning was ok.
If not, maybe one could insert an if statement: if(x> big_integer_number??) {throw something;}, to make sure the user is notified.

pboettch:
I think casting to double is not changing the behavior of the existing code: As we do `fabs()` on the next line which is implicitly casting to double precision is lost there, if there was anything to lose already.

While writing the lines above I realized that my code was already casting `x` implicitly to double. So your warning fix just puts things more explicit - which is great.

Nothing to worry.

eike-fokken:
Thanks for having a look!

PR #162, opened 2021-06-29
Remove conditional assignments to JSON_VALIDATOR_INSTALL

pdschubert:
Hi Patrick,

These two conditional assignments that turn `OFF` the `JSON_VALIDATOR_INSTALL` option do not seem to be correct. The assignments are causing an error in what I think is a relatively common use-case, which I describe in the following:

We are working on a project that uses nlohmann::json as a submodule. We would also like to use your json-schema-validator project as a submodule to validate json files that our project is handling. Our project uses CMake and, of course, also allows its users to install certain components to their system. When CMake installs the component of our project (as a library) that uses json-related code, it also needs to install nlohmann::json, which works fine, and it needs to install your json-schema-validator library as we are linking dynamically. But CMake throws an error at us when the `JSON_VALIDATOR_INSTALL` variable is set to `OFF`, since it has to install the library but does not find the respective `install(EXPORT ...` stuff that is hidden behind said variable. 

Removing those conditional assignments fixes the error and allows us to use your schema validator.

Thanks for this very useful project.

Best,
Philipp

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,7 +59,6 @@ if(TARGET nlohmann_json::nlohmann_json)
         nlohmann_json_schema_validator
         PUBLIC nlohmann_json::nlohmann_json)
 
-    set(JSON_VALIDATOR_INSTALL OFF)
     set(BUILD_TESTS OFF)
     set(BUILD_EXAMPLES OFF)
 
@@ -68,7 +67,7 @@ elseif(TARGET nlohmann_json) # or nlohmann_json, we are used a sub-project next 
     target_link_libraries(
         nlohmann_json_schema_validator
         PUBLIC nlohmann_json)
-    set(JSON_VALIDATOR_INSTALL OFF)
+
     set(BUILD_TESTS OFF)
     set(BUILD_EXAMPLES OFF)
 


Finkman:
If it is removed, does it make sense to keep the variable anyway?
It is once set to `ON` and never to `OFF` so the if-condition is always fulfilled.

pboettch:
I must say that I'm a bit fed with the complexity of this CMakeLists.txt . Used as a submodule, or has been installed, etc. Too many choices.

We should discuss how to simplify the CMakeLists.txt (and thus the integration possibilities) of this project.

Finkman:
In our projects, we always do `target_link_libraries` with `nlohmann_json::nlohmann_json`. The JSON lib provides that cmake target, either it is installed (via find_package) or as submodule (in-source).
That into account would boil down one branch in the cmake of the validator.

Next step, the validator could do similar to the approach in our projects, to do a `find_package` with `QUIET` and if that fails try an add_subdirectory. We do this in (almost) every project that has dependencies which allows to build them in-source and against installed libraries.

pboettch:
I remember that it is impossible to define a target with `::` in a normal CMakeLists.txt . So when using `nlohmann_json::nlohmann_json` it can't be with a submodule. Or am I wrong?

Finkman:
Take a look at JSON: https://github.com/nlohmann/json/blob/develop/CMakeLists.txt#L86
There, they create an alias `nlohmann_json::nlohmann_json` that points to `nlohmann_json`. So, they create a alias-target and they do it for exact the propose I mentioned.

PR #171, opened 2021-09-28
Switch default case was added

LLevella:
"error: switch missing default case [-Werror=switch-default]" was fixed


pboettch:
Thanks for contributing. What warning/error did you get? Which version of nlohmann::json were you using?

pboettch:
I'm not getting this error with gcc-10. It seems you're getting it because you explicitly activated `-Wswitch-default` in your environment. 

I prefer to keep it without a default-case because. This way I can catch real errors when a new type is added.

PR #172, opened 2021-09-28
Safe comparison with 0

LLevella:
error: comparing floating point with == or != is unsafe [-Werror=float-equal]


pboettch:
Which compiler (version) are you using to get this error?

pboettch:
There several comment out there saying that using `-Wfloat-equal` is misleading and considered harmful. E.g

https://github.com/fmtlib/fmt/issues/990#issuecomment-449831945

I won't fix this in this library and you should better not use this warning when compiling this (or even your code).

PR #173, opened 2021-09-30
Use explicit type conversion from json

qvfh83:
As detailed in https://github.com/nlohmann/json#implicit-conversions, it's recommended not to use implicit conversions from json. This change allows those using "JSON_ImplicitConversions" OFF in the nlohmann::json package to build this package without errors.

### DIFF
--- a/src/json-validator.cpp
+++ b/src/json-validator.cpp
@@ -715,13 +715,13 @@ public:
 	{
 		auto attr = sch.find("maxLength");
 		if (attr != sch.end()) {
-			maxLength_ = {true, attr.value().get<uint64_t>()};
+			maxLength_ = {true, attr.value().get<size_t>()};
 			sch.erase(attr);
 		}
 
 		attr = sch.find("minLength");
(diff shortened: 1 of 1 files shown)


pboettch:
I haven't got around making github actions yet for this project. Do all the unit-tests are still working?

qvfh83:
Yes, I ran them locally and they're all still working - if you really want I can try to run them within a containerized environment but may take a me a while to find time.

pboettch:
Great. Don't work on a local solution. If you have time it would be better to integrate github-actions to this project. There was once support for travis and we should do the same thing (compiling and running the unit-tests).

qvfh83:
@pboettch added actions - didn't quite use the travis template but I think it's ok (just squashed to single commit).

pboettch:
Thanks. Good idea to borrow from nlohmann! I think it'll need some changes however.

Could you make a dedicated PR addressing explicitly #174 ? Then I can integrate this one more easily.

Issue #181, opened 2021-11-17
CMake install regression


trevor-creator:
We are hoping to use nlohmann::json and json-schema-validator as submodules in our project. We are building for single-binary deployment, so our project is not meant to be installed using CMake. However, when I tried using master yesterday, I found that submodule use without installing no longer works. It appears that PR #165 has undone PR #81, re-causing error #79.

Example, using [the demonstration from #79](https://github.com/merlinmarek/json_schema_validator_issue) and starting with its original json and validator versions:

```
user@ubuntu:~/test/json_schema_validator_issue$ cmake -S . -B build
-- Using the single-header code from /home/user/test/json_schema_validator_issue/nlohmann_json/single_include/
-- Found nlohmann_json::nlohmann_json-target - linking with it
-- Configuring done
CMake Error: install(EXPORT "nlohmann_json_schema_validatorTargets" ...) includes target "nlohmann_json_schema_validator" which requires target "nlohmann_json" that is not in any export set.
-- Generating done
CMake Generate step failed.  Build files cannot be regenerated correctly.
user@ubuntu:~/test/json_schema_validator_issue$ cd nlohmann_json/
user@ubuntu:~/test/json_schema_validator_issue/nlohmann_json$ git checkout v3.10.4
Previous HEAD position was 7acf6218b :memo: fix typo #1903
HEAD is now at fec56a1a1 Merge branch 'release/3.10.4'
user@ubuntu:~/test/json_schema_validator_issue/nlohmann_json$ cd ../json_schema_validator/
user@ubuntu:~/test/json_schema_validator_issue/json_schema_validator$ git checkout  cc05de93274ee432ec75be5fbbd01e8d3b6d0ccd
Previous HEAD position was 8a7d1d3 Adapt CMake project name to be coherent with nlohmann::json's naming
HEAD is now at cc05de9 Create basic github-actions workflow (#175)
user@ubuntu:~/test/json_schema_validator_issue/json_schema_validator$ cd ..
user@ubuntu:~/test/json_schema_validator_issue$ cmake -S . -B build
-- The CXX compiler identification is GNU 9.3.0
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
[...]

PR #184, opened 2021-11-19
Optional CMake Install

trevor-creator:
Fix for #181 - option to not install re-added, but can be overridden. The default is to not install if included as a submodule, preserving old behavior. This is a branch off of #179.

Other changes:
* Install/don't install is now a single branch at the end of CMakeLists.txt.
* Moved the options below detection of top level project - previously, the options were overridden by the later SET statements, so setting them from the command line or a higher-level CMakeLists.txt wouldn't work.

Caveats: 
* If nlohmann::json and json-schema-validator are included as submodules, and json-schema-validator is set to install but nlohmann::json is not, the build will still fail.
* I've only tested this in the context of submodules, with various options set or not set. I haven't tested it as a standalone install.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,10 @@
-# Add Hunter support (Disabled by default)
-option(HUNTER_ENABLED "Enable Hunter package manager support" OFF)
-if(HUNTER_ENABLED)
+cmake_minimum_required(VERSION 3.2)
+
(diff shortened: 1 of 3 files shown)


pboettch:
All PRs have been closed due to me changing the main-branch to `main`. Please reissue your pull-request to that branch. Sorry and thanks.

Issue #189, opened 2021-12-10
Override default value of referenced schema


martin-koch:
Let us assume that we have a schema like this one:
```json
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "title": "A rectangle",
    "properties": {
        "width": {
            "$ref": "#/definitions/length",
	    "default": 20
        },
        "height": {
             "$ref": "#/definitions/length"
        }
    },
    "definitions": {
         "length": {
	      "type": "integer",
	       "default": 10
        }
    },
    "additionalProperties": false,
    "type": "object"
}
```
In my point of view I would expect that the default value of the concrete instance is considered instead of the default used by the referenced schema.
With the current implementation the validated and patched output of an empty json will be
```json
{"height":10,"width":10}
```
while I would have expected it to be
```json
{"height":10,"width":20}
```
I will suggest a change via a pull-request if you are okay with this.

pboettch:
I understand your approach. Overriding values "imported" by a `$ref` sounds like a good idea, but is forbidden by draft7:  https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-01#section-8.3 

> All other properties in a "$ref" object MUST be ignored.

Having said that, `default` is special as it has no meaning in the standard other then being there. Validation are free to use it in any sensible way - more or less.

Do you know whether in later json-schema-standards using $ref with overriding keywords is allow?

martin-koch:
It seems that is allowed since Draft2019. From the [current draft](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-8.2.3):
> Several keywords can be used to reference a schema which is to be applied to the current instance location. "$ref" and "$dynamicRef" are applicator keywords, applying the referenced schema to the instance.

There is also an [example](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-7.7.1.1) with overridden default values in the draft where it states:
> Feature A will therefore use a default value of true, while Feature B will use the generic default value of null. (...) Note that there are other reasonable approaches that a different application might take.  For example, an application may consider the presence of two different values for "default" to be an error, regardless of their schema locations.'

pboettch:
OK. Let's go see your PR.

PR #190, opened 2021-12-10
Support override of default value when using referenced schemas (#189)

martin-koch:
Solves #189 by copying the referenced schema and modifying the default value afterwards.


src/json-validator.cpp
```diff
@@ -1276,6 +1281,17 @@ std::shared_ptr<schema> schema::make(json &schema,
 			// so this is the origial URI for this reference, the $ref-value has thus be resolved from it
 			auto id = uris.back().derive(attr.value().get<std::string>());
 			sch = root->get_or_create_ref(id);
+
+			const auto defaultAttr = schema.find("default");
```
pboettch, line 1285: lower-case snake-case please
```suggestion
			const auto default_attr = schema.find("default");
```

src/json-validator.cpp
```diff
@@ -630,6 +630,11 @@ class type_schema : public schema
 			sch.erase(attr);
 		}
 	}
+
+	void setDefaultValue(const json &defaultValue)
```
pboettch, line 634:
```suggestion
	void set_default_value(const json &default_value)
```

test/issue-189-default-values.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+		// height must be 10 according to the default specified in the length definition while width must be 10 overridden by the width element
+		const auto expected = R"({"height":10,"width":20})"_json;
+		if (actual != expected) {
+			std::cerr << "Patch with defaults contains wrong value: '" << default_patch << "' instead of expected '" << expected.dump() << "'" << std::endl;
+			return 1;
+		}
+	}
+	return 0;
+}
```
pboettch, line 49: Could you add a test where two stacked/recursive `$refs` are used? And see whether it still works. I'm afraid that there is some late evaluation of not-yet loaded references which will break this mechanism.

pboettch (changes requested):

test/issue-189-default-values.cpp
martin-koch: Can you provide an example with stacked/recursive $refs and default values? I am not able to figure out a useful one...

pboettch: In addition to that, it should use other files to see whether late evaluation works. (You might laugh, but I don't remember whether I do late evaluation of references or not...) Let me check your branch and see what I can do.

pboettch: I adapted your example, it doesn't work if there is one more indirection. Could you check whether you can make it work. https://github.com/pboettch/json-schema-validator/tree/martin-koch/master

pboettch: OK, I'm looking at it now, I think there is no need to copy the schema. We just need to get the earliest default value. I'll try.

pboettch: Got it. I'll push on your master branch via this pull-request. Please test and review. And you're right, we need to copy the ref-schema when a (new) default is present.

pboettch: Damn it. There is a corner case regarding the copy of the ref... When the same ref on a ref is used, the default value won't be used... hmm, I don't understand myself, let's think about it a bit more.

pboettch: No, it's OK. The copy is actually what's need to be done. Thanks. If you're OK with these changes, could you update the README with this use-case and than I will happily merge it.

martin-koch: I updated the readme and fixed an issue with the simpler example (segmentation fault) which occurred after change 40f5d29 since get_or_create_ref may return an instance of schema_ref **or** an instance of type_schema.

martin-koch:
@pboettch are you okay with merging it?

pboettch:
Merged. Thank you for your patience.

pboettch:
All PRs have been closed due to me changing the main-branch to `main`. Please reissue your pull-request to that branch. Sorry and thanks.

Issue #198, opened 2022-04-20
Compatability issue with newest version of json


GKJose:
In json-validator.cpp, line 218, uri.pointer() != "" now throws an error.
Solution: uri.pointer().to_string() != ""

mingthemad:
Just found the same problem on Ubuntu 20.04.3 LTS using gcc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0

~/json-schema-validator/build$ make |& more
[  2%] Building CXX object CMakeFiles/nlohmann_json_schema_validator.dir/src/json-validator.cpp.o
/home/a_build/json-schema-validator/src/json-validator.cpp: In member function 'std::shared_ptr<{anonymous}::schema> nlohmann::json_schema::root_schema::get_or_create_ref(const nlohmann::json_uri&)':
/home/a_build/json-schema-validator/src/json-validator.cpp:219:21: error: no match for 'operator!=' (operand types are 'const json_pointer' {aka 'const nlohmann::json_pointer<std::__cxx11::basic_string<char> >'} and 'const char [1]')
  219 |   if (uri.pointer() != "") {
      |       ~~~~~~~~~~~~~ ^~ ~~
      |                  |     |
      |                  |     const char [1]
      |                  const json_pointer {aka const nlohmann::json_pointer<std::__cxx11::basic_string<char> >}
In file included from /home/a_build/json-schema-validator/src/nlohmann/json-schema.hpp:24,
                 from /home/a_build/json-schema-validator/src/json-validator.cpp:9:
/usr/local/include/nlohmann/json.hpp:13337:13: note: candidate: 'template<class RefStringTypeLhs, class RefStringTypeRhs> bool nlohmann::operator!=(const nlohmann::json_pointer<RefStringType>&, const nlohmann::json_pointer<RefStringTypeRhs>&)'
13337 | inline bool operator!=(json_pointer<RefStringTypeLhs> const& lhs,
      |             ^~~~~~~~
/usr/local/include/nlohmann/json.hpp:13337:13: note:   template argument deduction/substitution failed:
/home/a_build/json-schema-validator/src/json-validator.cpp:219:24: note:   mismatched types 'const nlohmann::json_pointer<RefStringTypeRhs>' and 'const char [1]'
  219 |   if (uri.pointer() != "") {
      |                        ^~
In file included from /usr/include/c++/9/regex:62,
                 from /home/a_build/json-schema-validator/src/json-validator.cpp:30:
[...]

PR #202, opened 2022-06-02
Fix explicit conversion to string

deimi:
Fixes https://github.com/pboettch/json-schema-validator/issues/198

Should also be backward compatible with nlohmann::json <3.11


deimi:
@pboettch If you wish I can also add a separate workflow for nlohmann::json 3.11

deimi:
@pboettch I've just seen now that nlohmann 3.11 is not released yet. So I suggest either to pull this change anyway as it is probably the cleaner and more explicit way to do it or wait till 3.11. is released and check if the incompatibility has remained.

Issue #210, opened 2022-07-04
Replace usage of dynamic_cast


inkychris:
`dynamic_cast` often requires RTTI to be enabled. If possible, it would be helpful to replace usage of `dynamic_cast` with some other mechanism so that this library can be used in environments that don't have RTTI.

SamVanheer:
I think this can be closed since #211 has been merged.

PR #211, opened 2022-07-05
Replace dynamic_cast with member functions (#210)

inkychris:
Replaces use of `dynamic_cast` with member functions on `schema`.

This is currently a minimal implementation, since most derived classes will return `nullptr`, but this is easy to remedy if the general solution is acceptable. This could be left as a pure virtual function in the `schema` class.

Both Issue 25 and 189 tests, and the JSON schema test suite pass locally. (Windows environment so not sure how tests are supposed to be run automatically).

Library now builds successfully without RTTI.

### DIFF
--- a/src/json-validator.cpp
+++ b/src/json-validator.cpp
@@ -40,6 +40,16 @@ protected:
 	root_schema *root_;
 	json default_value_ = nullptr;
 
+protected:
+	virtual std::shared_ptr<schema> make_for_default_(
+	    std::shared_ptr<::schema> & /* sch */,
+	    root_schema * /* root */,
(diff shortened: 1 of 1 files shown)


pboettch:
How does nlohmann-json handle dynamic_casts?

inkychris:
As far as I can tell, nlohmann-json doesn't use any dynamic casting. It's a bit of a heated topic but [some consider dynamic casting to be a code smell](https://stackoverflow.com/questions/48612271/use-case-of-dynamic-cast), i.e. sub-optimal design. My solution is by no means optimal, being a "just add a function to all the types" solution, but I suspect a more _proper_ solution would require a much larger change.

pboettch (approved):

Issue #212, opened 2022-08-19
📚️ What about "readOnly" ? (draft 7 section 10.3)


TrippyHank:
Hi there 👋

As of now, the validator does not honor [section 10.3](https://json-schema.org/draft-07/json-schema-validation.html#rfc.section.10.3) from the specification [JSON Schema Validation: A Vocabulary for Structural Validation of JSON](https://json-schema.org/draft-07/json-schema-validation.html#rfc.section.10.3).

I am interested in this excerpt specifically:

> If "readOnly" has a value of boolean true, it indicates that the value of the instance is managed exclusively by the owning authority, and attempts by an application to modify the value of this property are expected to be **ignored or rejected** by that owning authority.

It is clearly stated that attempts by an application to change the value of this property must be **ignored or rejected**.

I was expecting one of these two options for validation:
1. Ignoring properties for which the schema defines a "readOnly" key with a value of true
2. Throwing an exception informing the user that the value for this key is read-only

pboettch:
I think the key to understand how read-only works is the phrase "change the value of this property". When does a value change? Clearly not during validation. 

It seems to me that this is not validator's work. The validator is not the "owning authority". It is your application.

But maybe the validator can be used as an instance or frontend for your application to apply this functionality.

TrippyHank:
This library encapsulates everything needed to honor this section easily.
The user does not have any clue about all the objects that composes the specified schema tree behind the scene, if there was access i might agree with you.
The user would have to rewrite some parts of this library to identify the `readOnly` nodes, which is not really ideal.

TrippyHank:
It certainly doesn't cover everything from [Schema Annotations](https://json-schema.org/draft-07/json-schema-validation.html#rfc.section.10), but it does the trick for now 🍻

pboettch:
As I wrote in the PR: the validator could extract and provide a list of json-paths which you could use in your applicaiton to handle the read and write-Only fields. What do you think?

TrippyHank:
As written in #213, I think that
> it's an attractive idea too but it adds an unnecessary burden to the user from my point of view

pboettch:
Coming back to your original idea: could you create a (pseudo)-code snippet of how you would like to use this feature as a user with the validator? Here's my vision:

This schema
```JSON
{
  "properties": {
    "foo": { "type": "int", "readOnly": true},
    "bar": { "type": "int", "writeOnly": true},
    "baz": { "type": "int"}
  }
}
```

[...]

PR #213, opened 2022-08-19
📚️ raise error when trying to change a read only value

TrippyHank:
Relates to #212 

This pull request honors "readOnly" keyword from [section 10.03](https://json-schema.org/draft-07/json-schema-validation.html#rfc.section.10.3) from the [specification](https://json-schema.org/draft-07/json-schema-validation.html) by raising an error 🐢

### DIFF
--- a/src/json-validator.cpp
+++ b/src/json-validator.cpp
@@ -457,6 +457,7 @@ class type_schema : public schema
 	std::vector<std::shared_ptr<schema>> type_;
 	std::pair<bool, json> enum_, const_;
 	std::vector<std::shared_ptr<schema>> logic_;
+	bool read_only_ = false;
 
 	static std::shared_ptr<schema> make(json &schema,
 	                                    json::value_t type,
@@ -506,6 +507,13 @@ class type_schema : public schema
 				if (else_)
 					else_->validate(ptr, instance, patch, e);
 			}
+		}
+
+    // enforce default value if any read only values
+		if(read_only_) {
+      auto rod_value = default_value(ptr, instance, e);
+      if(rod_value != nullptr && instance != rod_value)
+        e.error(ptr, instance, "key is read-only");
 		}
 	}
 
@@ -559,6 +567,12 @@ public:
 		attr = sch.find("default");
 		if (attr != sch.end()) {
 			set_default_value(attr.value());
+			sch.erase(attr);
+		}
+
+		attr = sch.find("readOnly");
+		if (attr != sch.end()) {
+			read_only_ = attr.value().get<bool>();
 			sch.erase(attr);
 		}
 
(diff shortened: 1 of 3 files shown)


pboettch:
Are you sure that this is how read-only shall work? If the value is not equal to the default value (if any) it shall fail?

Isn't read-only only for run-time usage of an schema-validated instance? So out-of-scope of this library?

pboettch:
In any case, we need a test before integrating changes. I think in this case the tests will also show the limitations of this approach.

TrippyHank:
Since this library allows `defaults` parsing (which is awesome), it should reflect the referenced section of the specification in the process

pboettch:
readOnly has nothing to do with default-values. I don't understand. I read the section and I strongly think that readOnly is something to be ignore by the validator. I could be exposed as the validator compiles the schema but the validation-process itself should ignore it.

It has to be handled by the owning authority.

TrippyHank:
The validator has ownership over the objects composing the schema tree, it should enforce the default value.

> If "readOnly" has a value of boolean true, it indicates that the value of the instance is managed exclusively by the owning authority

The **owning authority** is the validator

> and attempts by an application to modify the value of this property are expected to be ignored or rejected by that owning authority

Since the validator owns the objects that make up the schema tree, it must **ignore or reject** attempts to change the value of this key in the patch resulting from the validation process at run time.

TrippyHank:
Do you think it should also rejects attempts when the `key` has no `default` ?

pboettch:
> The **owning authority** is the validator since it owns the instances that make up the schema tree.

False, the schema validator gets the to-be-validated instance as a `const`. How can it own something which can't be modified? The whole instance is read-only from the validator's point of view.

If you're still in doubt about that, please check with the json-schema-project to clarify.  https://github.com/json-schema-org/community/discussions

pboettch:
Closing for the moment. Please re-open once you have a feedback showing that I'm wrong.

TrippyHank:
### Let's begin the journey to _"Why are you wrong?"_

> the schema validator gets the to-be-validated instance as a const. 

This is absolutely true, we both agree! The user-defined input should stay _read only_ 😉

> How can it own something which can't be modified? The whole instance is read-only from the validator's point of view.

There is obviously a misunderstanding here since we agree again, so I will do my very best to clarify.

Users are writing `applications` including the `library`.

1. They initialize a `validator` with a `root schema` that they crafted themselves.
2. They use this `validator` to validate a user-defined `instance` against the `root schema` embedded by the `validator`.
3. They merge a copy of the `default instance` of the `root schema` with the user-defined `instance` if it passes validation.

Questions:
- How is the user supposed to get rid of the `readOnly` keys in **step 3**?
- Do you agree that since objects composing the schema are created at the validation stage (when `.validate(instance)` is called), the validator is the owner of these objects?
- If you disagree with the previous question, who owns these objects?

With all this, if you still don't see why this is the way to go, I can't do anything more for this library

> If you're still in doubt about that, please check with the json-schema-project to clarify. https://github.com/json-schema-org/community/discussions

I'll leave it as is, hopefully if your opinion evolves or if you find tests that fail, tag me 🍻

> However, what the validator could do is to provided a list of readOnly JSON-paths which you can used in your application to check when the user changes something in the instance.

I thought about a vector of `json pointer`, it's an attractive idea too But it adds an unnecessary burden to the user from my point of view

PR #214, opened 2022-08-23
Update CI and fix warnings

res0nance:
Hi @pboettch, PTAL.

This updates CI to run on main and uses the latest nlohmann-json to test.
This also adds the `to_string()` call to remove warnings.

The current tag is also very old so could we get an updated tag?

### DIFF
--- a/.github/workflows/github-actions.yml
+++ b/.github/workflows/github-actions.yml
@@ -37,3 +37,30 @@ jobs:
         run: cmake --build build --target all -j$(nproc)
       - name: test
         run: cd build && ctest
+  build_and_test_min_version:
+    runs-on: ubuntu-latest
+    container: ghcr.io/nlohmann/json-ci:v1.0.0
+    steps:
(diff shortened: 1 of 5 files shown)


pboettch:
If we test with 3.11 should we also limit the version use in the code?

res0nance:
> If we test with 3.11 should we also limit the version use in the code?

I don't think we have to these changes should be backwards compatible since the use of `to_string()` already appears in this library and in nlohmann/json, It claims that the addition of `to_string()` has been there since 2.0.0

I think what we should do is perhaps add another CI job that tests on the oldest supported version which is 3.8.0.

pboettch:
That sounds like a good idea.

When you say the tag is very, what you mean is you'd like to have a new(er) release?

res0nance:
> That sounds like a good idea.
> 
> When you say the tag is very, what you mean is you'd like to have a new(er) release?

Yeah a new release would be good considering that the last one was 2 years ago.

pboettch:
Time goes by very fast.

Issue #218, opened 2022-09-02
Validation of schema using $ref always returns true 


nine:
**Observed behavior**

The validation of following Json schema using the `$ref` keyword always returns true.

```
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "title": "Schedule",
    "type": "object",
    "properties": {
        "reply": {
            "description": "Send reply",
            "type": "boolean"
        },
        "schedule": {
            "description": "Array with schedule entries",
            "type": "array",
            "items": { "$ref": "#/definitions/schedule_entry" }
        }
    },
    "definitions": {
        "schedule_entry": {
            "type": "object",
            "required": [
                "id",
                "startdate",
                "cmd",
                "settings"
            ],
            "properties": {
                "id": {
                    "type": "string"
                },
                "startdate": {
                    "type": "number",
                    "minimum": 0
                },
                "cmd": {
                    "type": "string"
                },
                "settings": {
                    "type": "object"
                }
            }
        }
    }
}
```

Tested with 
1. valid object:
   ```
   {
    "reply": false, 
    "schedule": [
        {
            "id": "abc",
            "startdate": 1,
            "cmd": "foo",
            "settings": {}
        }
    ]
   }
   ```
2. invalid object
   ```
   { "foo": "bar" }
   ```

**Expected behavior**

The validation of the second Json instance should return false.

**Steps to reproduce**
A unit test for the bug is available in #219.

nine:
Discovered missing "required" array that caused wrong expectations on validation behavior. json-schema-validator works as expected!

PR #219, opened 2022-09-02
Adds test for issue #218

nine:
Adds test to reproduce issue #218


nine:
Discovered missing "required" array that caused wrong expectations on validation behavior. json-schema-validator works as expected!

Issue #152, opened 2021-03-19
email and idn-email validation not based on RFC-5321, as it should be


gene-hightower:
The JSON schema validation spec defines valid idn-email and email addresses in terms of RFC-5321.
<https://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.7.3.2>

The RFC-5321 grammar is regular (Chomsky's type-3), as opposed to the grammar of RFC-5322 which is *recursive* (type-0) thus requiring a more powerful tool. That is to say, the 5321 grammar can be recognized using a regular expression.

pboettch:
Thanks for this information. This library only support draft-7 and IIRC there it is not mandatory to have a validator for email-addresses. Everything is optional. But this library has the `format_checker`-callback which allows you to validate anything which has the format-field-set in your custom code.

gene-hightower:
Let me offer what I think is a very small, fast, and (I hope) correct function to validate an RFC-5321 Mailbox address.
<https://github.com/gene-hightower/smtp-address-validator>
It uses a state machine generated by <https://github.com/adrian-thurston/ragel> -- this tool is available to install as a package on Fedora and Ubuntu, and I suspect most major platforms.
This function compiles to about 10K on x86_64 and allocates no dynamic memory.

gene-hightower:
What do you need to use smtp-address-validator in support of idn-email and email validation?
Would a package of some type for smtp-address-validator help? What type of package?

pboettch:
I think, if we integrate a third-party library, it should be copied here, if it not too big. So what is too big, I don't know. I'm open to ideas.

gene-hightower:
Not too big should describe two files, 770 lines of code.
With this implementation, you should be able to pass tests for email and idn-email in draft2019-09 and draft2020-12.

PR #235, opened 2022-10-21
Use Ragel generated RFC-5321/RFC-6531 Mailbox address parser

gene-hightower:
Generated code copied from: <https://github.com/gene-hightower/smtp-address-parser>

fixes: #152


pboettch:
Could you try to fix the errors the CI reported?

gene-hightower:
> Could you try to fix the errors the CI reported?

Oh, right. I've used C++17 without even really thinking about it.

gene-hightower:
So, I'm proposing to require C++17, as it's the default version of C++ for GCC and Clang now if you don't specify something else, it's been the standard for over five years, and it's almost fully supported by the major compilers.

pboettch:
Sorry for the late response.

I want to stick to nlohmann::json's requirements and this is C++11. Is it possible for you to change your code to make it work with only C++11 - features?

gene-hightower:
> Is it possible for you to change your code to make it work with only C++11 - features?

Sure.

pboettch (approved):

pboettch:
I rebased your changes and integrated them in a squashed commit.

PR #240, opened 2022-11-28
Local include for smtp-address-validator.hpp

garethsb:
This makes _string-format-check.cpp_ consistent with _smtp-address-validator.cpp_.
This means that the _nlohmann_ directory itself doesn't have to be in the system search path. (The directory in which the _nlohmann_ directory is contained is much more likely to be, so e.g. `#include <nlohmann/json-schema.hpp>` is fine.)

### DIFF
--- a/src/string-format-check.cpp
+++ b/src/string-format-check.cpp
@@ -1,6 +1,6 @@
 #include <nlohmann/json-schema.hpp>
 
-#include <smtp-address-validator.hpp>
+#include "smtp-address-validator.hpp"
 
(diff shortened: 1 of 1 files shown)


pboettch:
Yes, I saw that after committing. Do you think this can generate build problems somewhere?

garethsb:
Did for me after copying files into same source tree where I was using version 2.1.0 previously. Easily fixed 🙂

pboettch:
OK, then I will wait for some time before doing a 2.2.1 .

PR #242, opened 2022-11-30
Fix assumed signed char

bobolopolis:
The code assumes that char is signed, but whether char is signed or unsigned is implementation defined. On some architectures like PowerPC, GCC treats char as unsigned resulting in compile errors:

	smtp-address-validator.cpp:213:1: error: narrowing conversion of '-32' from 'int' to 'char' [-Wnarrowing]

Fix this by specifying signed char.


garethsb:
This (122 errors) also occurs on my GCC 9.4 build on linux/arm64, so thanks for the fix.

PR #276, opened 2023-06-15
Fix bug: attr.value() of an array attribute returns an iterator of basic_json objects not an iterator of std::string

VolkerChristian:
This fixes also

### DIFF
--- a/src/json-validator.cpp
+++ b/src/json-validator.cpp
@@ -604,10 +604,12 @@ public:
 			} break;
 
 			case json::value_t::array: // "type": ["type1", "type2"]
-				for (auto &schema_type : attr.value())
+				for (auto &array_value : attr.value()) {
+					auto schema_type = array_value.get<std::string>();
 					for (auto &t : schema_types)
 						if (t.first == schema_type)
 							type_[static_cast<uint8_t>(t.second)] = type_schema::make(sch, t.second, root, uris, known_keywords);
+				}
 				break;
 
 			default:


pboettch:
Should we add a new json version to the CI?

VolkerChristian:
Sorry, i din't get the question.

pboettch:
Does this problem is coming from new nlohmann::json version or from a new GCC version? In any case, this needs to be tested with an github-action. Would you be able to update your PR so that it tests the new version, respectively?

VolkerChristian:
OK, now I understand.
This problem is due to a new GCC version 13.1.0 and it obviously manifests itself also when using the latest MSVC 17.6.2.
I think the compilers have become stricter with respect to template argument deduction. To be honest, I have not enough time to look deeper at what changes in 13.1.0 triggers this bug now. This does not happen with gcc version 12.3.0 and clang version 14.0.6. I just tested with the latest stable release of clang (16.0.6) which also _didn't_ trigger that bug.

- gcc-12.3.0: OK
- clang-14.0.6: OK
- clang-16.0.6: OK
- gcc-13.1.0: NOK 
- MSVC 17.6.2: NOK (not tested by me)

But do you really need a test for that? And what explicitly should be tested?

As the documentation of `nlohman::json` explicitly noted, that `array.value()` returns an iterator which can be used in a range based for loop to access all json-objects (values) of the array. And because the concrete array is an array of json type names (as `std::strings)` a concrete type name has to be accessed using `.get<std::string>().`

xamix:
The bug raised with Clang-16.0.6 for me:

```
(Ubuntu clang version 16.0.6 (++20230610113410+7cbf1a259152-1~exp1~20230610233423.102)
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

Not with Clang-15:
```
Ubuntu clang version 15.0.7
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

Anyway I will test this fix
Regards.

Issue #229, opened 2022-09-27
Default values of all "oneOf" sub-schema are inserted in patch 


didier-brizet:
Only the default value of one of "oneOf" sub-schema must be inserted in patch returned by `json_validator::validate`.

For instance, the validation of ``` {"name": "bar"} ``` against the following JSON schema:
```
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "type": "object",
    "oneOf": [
        {
            "type": "object",
            "properties": {
                "name": {
                    "enum": "foo"
                },
                "code": {
                    "const": 1,
                    "default": 1
                }
            }
        },
        {
            "type": "object",
            "properties": {
                "name": {
                    "enum": "bar"
                },
                "code": {
                    "const": 2,
                    "default": 2
                }
            }
        }
    ]
}
````

must returns the patch ```[{"op":"add","path":"/code","value":2}]``` but the following patch is returned:
```[{"op":"add","path":"/code","value":1},{"op":"add","path":"/code","value":2}]```

didier-brizet:
I propose the following patch:

[issue_229.txt](https://github.com/pboettch/json-schema-validator/files/9656902/issue_229.txt)

UX3D-becher:
@pboettch We have the same issue with a complex schema, is there anything that still needs to be done in the PR before it can be merged?

pboettch:
It's a huge change. Hard to review. Especially if you're not convinced of the handling of default values _inside_ the validator.

I'd rather prefer to have an outside-of-the-validator solution of default-value-handling. But this is off-topic for this PR.

pboettch:
I won't accept big changes anymore on patch creation concerning default values. I'm not convinced that this library is the right place. Or at least it should not be integrated in the code as it is today.

PR #288, opened 2023-07-20
Fix performance regression from logical combination patch discard

sjoubert:
A recent fix to discard patch from invalid logical combination introduced a lot (for large json instance and/or schema) of copies to make a patch backup. On some case, this introduced up to a 100x slower validation time (I observed times from 50ms to 7s on one of my use case).

Resizing the patch object to its previous size avoid unnecessary temporary allocations from the backup object.

* I needed to expose the internal json object to be able to resize it. I'm not sure how differently we could achieve this if you don't like that.
* It also fixes the default value of an empty json_patch. I think it should be an empty array and not a null json according to the json patch schema.
* I know you mentioned in #229 that you won't accept more changes on this subject, however this is quite a big performance regression so I hope this can still be merged.
On that matter, I'm willing to spend a bit of time, after this fix, on the subject if you have ideas on how to better integrate the default values handling in the library. I think this is quite an important feature as it makes parsing a lot easier if one can put default values in a single reference place (the schema). I don't known what elements you don't like in the current design though.

### DIFF
--- a/src/json-patch.hpp
+++ b/src/json-patch.hpp
@@ -28,10 +28,13 @@ public:
 	json_patch &replace(const json::json_pointer &, json value);
 	json_patch &remove(const json::json_pointer &);
 
+	json &get_json() { return j_; }
+	const json &get_json() const { return j_; }
+
 	operator json() const { return j_; }
 
 private:
-	json j_;
+	json j_ = nlohmann::json::array();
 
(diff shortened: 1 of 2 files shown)


pboettch:
Sorry to hear, indeed: there are not performance regression tests.

PR #290, opened 2023-08-17
Fix Clang compiler warnings

serge-s:
Fix "unused variable" warnings generated by Clang.
This fix can be useful for those who want to build with -Werror


serge-s:
"test / Run ctests" actions failure doesn't seem to be related to these changes - they all fail at CMake configure stage while I'm only touching C++ code.

PR #296, opened 2023-09-21
use cmake 3.24 for runners

pboettch:


pboettch:
Well, python3-pip doesn't exist in my runners.

JacobCrabill:
Try adding `apt update && apt upgrade -y` before the call to `apt install -y python3-pip`; I think the container just needs to collect the list of available packages first.

PR #299, opened 2023-10-12
Added json-patch validation 'remove' info

Oplkill:
In json validation added json-path info for keys that didn't described in json-schema as 'remove'.


pboettch:
Could you please elaborate more about what this fixes, and add a test to the test-suite which tests this feature.

PR #343, opened 2025-01-28
WIP converting all static strings over to draft 2020-12

gmabey:


src/json-validator.cpp
```diff
@@ -1409,7 +1409,7 @@ std::shared_ptr<schema> schema::make(json &schema,
 
 			schema.erase(attr);
 
-			// special case where we break draft-7 and allow overriding of properties when a $ref is used
+			// special case where we break draft-7 and allow overriding of properties when a $ref is used  <<-- might be out-of-date now that 2020-12 is used
```
gmabey, line 1412: Can someone help me understand the meaning in this comment?

src/nlohmann/json-schema.hpp
```diff
@@ -37,7 +37,7 @@ namespace nlohmann
 
 // A class representing a JSON-URI for schemas derived from
 // section 8 of JSON Schema: A Media Type for Describing JSON Documents
-// draft-wright-json-schema-00
+// draft-wright-json-schema-00  <<-- might be incorrect now that 2020-12 is used
```
gmabey, line 40: same as above

src/nlohmann/json-schema.hpp
```diff
@@ -128,7 +128,7 @@ class JSON_SCHEMA_VALIDATOR_API json_uri
 namespace json_schema
 {
 
-extern json draft7_schema_builtin;
+extern json draft_2020_12_schema_builtin;
```
gmabey, line 131: I left this general approach here, because I suspect that the intent is to allow for multiple versions of the schema to be included simultaneously -- and validate against the one indicated by `$schema`. Is that correct?
If not, then this variable should be renamed to something more generic, like `schema_builtin`.

test/JSON-Schema-Test-Suite/tests/draft_2020-12/definitions.json
```diff
@@ -1,12 +1,12 @@
...
     {
         "description": "validate definition against metaschema",
-        "schema": {"$ref": "http://json-schema.org/draft-07/schema#"},
+        "schema": {"$schema": "http://json-schema.org/draft/2020-12/schema"},
         "tests": [
             {
                 "description": "valid definition schema",
                 "data": {
-                    "definitions": {
+                    "$defs": {
```
gmabey, line 9: I don't think this is correct; the test is still failing (er, passing when it should fail). Could someone help me understand the structure here better?

test/JSON-Schema-Test-Suite/tests/draft_2020-12/definitions.json
```diff
@@ -15,7 +15,7 @@
             {
                 "description": "invalid definition schema",
                 "data": {
-                    "definitions": {
+                    "$defs": {
```
gmabey, line 18: same as above.

test/JSON-Schema-Test-Suite/tests/draft_2020-12/definitions.json
rpatters1: I have barely stuck my toe into json schema definitions, so it is entirely possible that your question is mostly over my head. But draft 2020-12 allows either `definitions` or `$defs` as a keyword. Is that what you were expecting to fail? Because if so, it should, in fact, pass. My PR #338 allows either to pass precisely for this reason.

gmabey: I think I wrote that comment when I was still trying to get my bearings on stuff.
I think you'll agree that `$defs` is "correct", which `definitions` is the backward compatible safety net, no?
At least, according to https://json-schema.org/draft/2020-12/schema

rpatters1: >I think you'll agree that $defs is "correct", which definitions is the backward compatible safety net, no?

That's the way I read it as well.

PR #370, opened 2025-12-24
Fix crash when validating null values in objects with additionalProperties

orgads:
Remove erroneous code that added a patch with empty JSON pointer for any null instance. Null is a valid JSON value and should not trigger a patch.

### DIFF
--- a/src/json-validator.cpp
+++ b/src/json-validator.cpp
@@ -588,9 +588,6 @@ class type_schema : public schema
 					else_->validate(ptr, instance, patch, e);
 			}
 		}
-		if (instance.is_null()) {
-			patch.add(nlohmann::json::json_pointer{}, default_value_);
-		}
 	}
 
 protected:
--- a/test/issue-null-values-in-object.cpp
+++ b/test/issue-null-values-in-object.cpp
@@ -1,0 +1,33 @@
+#include <iostream>
+#include <nlohmann/json-schema.hpp>
+
+using nlohmann::json;
+using nlohmann::json_schema::json_validator;
+
+// Test: object with additionalProperties containing null value should validate
+// additionalProperties causes traversal into the object where null value triggers the bug
+static const json schema = R"(
+{
+    "properties": {
+        "x": {"type": "object", "additionalProperties": {}}
+    }
+})"_json;
+
+int main(void)
+{
+	json_validator validator{};
+	validator.set_root_schema(schema);
+
+	json instance = R"({"x": {"nested": null}})"_json;
+
+	const auto patch = validator.validate(instance);
+	const auto result = instance.patch(patch);
+
+	// null value should be preserved
+	if (!result["x"]["nested"].is_null()) {
+		std::cerr << "Failed: nested should be null, got: " << result["x"]["nested"].dump() << std::endl;
+		return 1;
+	}
+
+	return 0;
+}
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -96,3 +96,7 @@ add_test(NAME issue-255-error-message-limit-precision COMMAND issue-255-error-me
 add_executable(issue-105-verbose-combination-errors issue-105-verbose-combination-errors.cpp)
 target_link_libraries(issue-105-verbose-combination-errors nlohmann_json_schema_validator)
 add_test(NAME issue-105-verbose-combination-errors COMMAND issue-105-verbose-combination-errors)
+
+add_executable(issue-null-values-in-object issue-null-values-in-object.cpp)
+target_link_libraries(issue-null-values-in-object nlohmann_json_schema_validator)
+add_test(NAME issue-null-values-in-object COMMAND issue-null-values-in-object)


test/issue-null-values-in-object.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+using nlohmann::json;
+using nlohmann::json_schema::json_validator;
+
+// Test: object with additionalProperties containing null value should validate
+// additionalProperties causes traversal into the object where null value triggers the bug
+static const json schema = R"(
+{
+    "properties": {
+        "x": {"type": "object", "additionalProperties": {}}
```
pboettch, line 12: additionalProperties is defaulting to be an empty schema. So can be removed.

orgads: I'm not sure why, but if I remove this, the test succeeds even without the fix.

src/json-validator.cpp
```diff
@@ -588,9 +588,6 @@ class type_schema : public schema
 					else_->validate(ptr, instance, patch, e);
 			}
 		}
-		if (instance.is_null()) {
-			patch.add(nlohmann::json::json_pointer{}, default_value_);
-		}
```
pboettch, line 593: I see the point, but what is then the condition to apply the default value to the patch?

test/issue-null-values-in-object.cpp
pboettch: Empty schema is seen as maybe seen as false. So additionalProperties are not allowed.

Not sure whether this is expected.

src/json-validator.cpp
orgads, line 1101: It is done here, isn't it? Not sure why it was added above in the first place. All the tests pass.

src/json-validator.cpp
orgads: See below on line 1101.

test/issue-null-values-in-object.cpp
pboettch: Definitely something wrong here, `additionalProperties: {}` should behave the same as if omitted.

orgads: 🤷‍♂️

Without the fix:
* Without `additionalProperties`, the test passes
* With `additionalProperties`, I get this:
```
ctest -R issue-null --output-on-failure
Test project /home/orgads/Projects/json-schema-validator/build
    Start 18: issue-null-values-in-object
1/1 Test #18: issue-null-values-in-object ......Subprocess aborted***Exception:   0.02 sec
terminate called after throwing an instance of 'nlohmann::json_abi_v3_12_0::detail::type_error'
  what():  [json.exception.type_error.305] cannot use operator[] with a string argument with null


0% tests passed, 1 tests failed out of 1
```

pboettch: It's late, but I read a little bit further: The default-patch handling seems flawed: 

```C++
		if (instance.is_null()) {
			patch.add(nlohmann::json::json_pointer{}, default_value_);
		}
```
I don't understand how the, seemingly, global instance patch, can use the root-pointer. Such a patch, when applied, will overwrite everything from the instance.

This is what you see in your test.

The incoherence of your test result regarding `additionalProperties` (present as `{}` or omitted) comes from the fact that nested is a null value, and when `additionalProperties` is `{}` it enters validation nested, this succeeds, but as it is a null-value a patch is added, with in my eyes a wrong json-path.

I think removing the if as you did, or at least change it: 

```C++
if (instance.is_null() && default_value_) {
	patch.add(ptr, default_value_);
}
```

This would still create a patch for a possible default value of the instance is null. So this is wrong.

Actually a default value would need to be added only if the json-path does not exist in the validated instance.

As it is stands, default values patch creation seems flawed...

Not sure what to do, is the original author still available to check and explain? @Finkman 

This change 59c9d6200bf3cd54b4fc717ec1660c91eddb4d1a was adding these lines.

pjungkamp: The change https://github.com/pboettch/json-schema-validator/commit/59c9d6200bf3cd54b4fc717ec1660c91eddb4d1a is part of https://github.com/pboettch/json-schema-validator/pull/244 and was introduced to fix https://github.com/pboettch/json-schema-validator/issues/243.

I think that the issue, the PR, and the test case are fundamentally wrong here.

Take a look at the example schema from the issue:

```json
{
    "properties": {
        "huhu": {
            "type": "number"
        }
    },
    "type": "object",
    "default":{
      "huhu": 4
    }
}
```

The author expects this schema to validate successfully for a `null` value and return a non-empty patch. The expected behavior here is obviously to reject the `null` value because it doesn't match the `"type": "object"` predicate in the schema.

The [test case](https://github.com/pboettch/json-schema-validator/blob/main/test/issue-243-root-default-values.cpp) added in the PR makes that misunderstanding pretty clear and omits the `"type": "object"` predicate for the test case to be able to succeed.

I think that PR/commit just needs to be reverted, as it is just wrong.

PR #371, opened 2025-12-24
Fix required property error when property has a default value

orgads:
Skip the 'required property not found' error if the property has a default value defined, since the default will be added to the patch by the existing reverse search loop.


pboettch:
As much as I like the cleanliness of your implementation, as the default field is not a requirement of the standard, standard-compliant validation has to fail, even if a default value is provided.

At least that is my understanding.

Can you prove me otherwise?

orgads:
You're probably right. ajv does accept required fields with default values, but I also tester with another validator (I think blaze), and it didn't so I guess this can be closed.

PR #376, opened 2026-05-21
Fix is_address when char is unsigned

### LINKED ISSUE #375: is_address can return incorrect results when char is unsigned
On platforms where the char type is unsigned, such as aarch64, one test fails:
```
67/95 Test #67: JSON-Suite::Optional::Format::idn-email ...............***Failed    0.00 sec
Testing Group "validation of an internationalized e-mail addresses"
  Testing Case "a valid idn e-mail (example@example.test in Hangul)"
    Test Case Exception (invalid argument): At  of "실례@실례.테스트" - format-checking failed: 실례@실례.테스트 is not a valid idn-email according to RFC 6531.
      --> Test Case exited with 0 NOT expected.
  Testing Case "an invalid idn e-mail address"
    Test Case Exception (invalid argument): At  of "2962" - format-checking failed: 2962 is not a valid idn-email according to RFC 6531.
      --> Test Case exited with 0 as expected.
  Testing Case "a valid e-mail address"
      --> Test Case exited with 1 as expected.
  Testing Case "an invalid e-mail address"
    Test Case Exception (invalid argument): At  of "2962" - format-checking failed: 2962 is not a valid idn-email according to RFC 6531.
      --> Test Case exited with 0 as expected.
Group RESULT: "validation of an internationalized e-mail addresses" 3 of 4 have succeeded - 1 failed
-------------
Total RESULT: 3 of 4 have succeeded - 1 failed
```

jamesjer:
Fixes https://github.com/pboettch/json-schema-validator/issues/375.  The issue is that, in the comparison `if (((*(p))) < (*(_mid)))`, `p` points to a `char` and `_mid` points to a `signed char`.  If `_mid` is negative, then the comparison is always false.  This PR makes all pointers explicitly point to signed values to fix the issue.

### DIFF
--- a/src/smtp-address-validator.cpp
+++ b/src/smtp-address-validator.cpp
@@ -670,11 +670,13 @@ static const short _address_eof_trans[] = {
 
 static const int address_start = 1;
 
-bool is_address(const char *p, const char *pe)
+bool is_address(const char *pc, const char *pec)
 {
 	int cs = 0;
 
-	const char *eof = pe;
+	const signed char *p = reinterpret_cast<const signed char *>(pc);
+	const signed char *pe = reinterpret_cast<const signed char *>(pec);
+	const signed char *eof = pe;
 
 	bool result = false;
 


sudipm-mukherjee:
Thanks for the PR @jamesjer, I have added this patch to Debian package now and is fixing the issue I reported at https://github.com/pboettch/json-schema-validator/issues/377

PR #383, opened 2026-08-10
Fix schema edgecases

pjungkamp:
## Description

I've taken my time to look deeper into this schema-validator and noticed some edge-cases in the default value processing and error reporting.

### Multiple validation errors for "additionalProperties"

I commonly use the `additionalProperties` directive to specify a schema for a string-value-map-style objects. The validator will only output the first validation error encountered in any validated property. I'd compare this case to the validation of array items where we are not limited to a single error.

Let's assume the following schema:

```json
{
  "type": "object",
  "additionalProperties": {
    "type": "array",
    "items": [
      { "const": "A" },
      { "const": "B" }
    ] 
  }
}
```

If we now validate e.g. `{ "some_key": [1, 2] }` we will only see an error reported for the first element of the array. The diagnostic for the second element is discarded by the validator.

### Failed conditions can insert default values

Condition-type schema directives like `if`, `not` and `contains` which apply a subschema that is expected to fail even if validation ulimately succeeds need to ensure that their subschema will not introduce default values into the default value patch. I've now implemented this in a way that disregards default values even if the schema validates successfully for `if`, `not` and `contains` in the same way that was already done for `anyOf` and `oneOf`.

### Error messages for trivial false-schemas

A common pattern used to deny unknown properties on object schemas or to deny unknown additional items in tuple-style array schemas is to set `additionalProperties` or `additionalItems` to `false`. This indicates that only a closed-set of indices or properties are accepted here.

The current message of `validation failed for additional property '...': instance invalid as per false-schema` doesn't quite sit right with me, so I'm now reporting `found unexpected property "..." in object` or `found unexpected index ... in array`.

### Broken default values

I'd also like to refer to #370 which fixes an urgent problem with the way default values are processed in this library.

### DIFF
--- a/src/json-patch.cpp
+++ b/src/json-patch.cpp
@@ -101,6 +101,14 @@ json_patch &json_patch::remove(const json::json_pointer &ptr)
 	return *this;
 }
 
+json_patch::saved_size json_patch::save() const {
+	return saved_size{j_.get_ref<json::array_t const &>().size()};
+}
+
+void json_patch::restore(saved_size size) {
+	j_.get_ref<json::array_t &>().resize(static_cast<json::array_t::size_type>(size));
+}
+
 void json_patch::validateJsonPatch(json const &patch)
 {
 	// static put here to have it created at the first usage of validateJsonPatch
--- a/src/json-patch.hpp
+++ b/src/json-patch.hpp
@@ -20,6 +20,8 @@ private:
 class json_patch
 {
 public:
+	enum class saved_size : json::array_t::size_type {};
+
 	json_patch() = default;
 	json_patch(json &&patch);
 	json_patch(const json &patch);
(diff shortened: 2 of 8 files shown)


pboettch:
For all of your commits we need a test-case showing what's wrong and what got fixed. I'm having a hard-time understanding the problem. TBH, this looks like an AI-based PR (excuses if it is not), which is no problem "en soit", but it needs more human touch to be processed.

pjungkamp:
No AI was used here at all. I'll add test cases. 👍

pboettch:
Could you check how your MR matches or is superseded by #384  ? Could it make your changes even easier?

pjungkamp:
I really wish that commit 1 could just be merged into that PR instead.

Commit 2 addresses a real problem where schema validation produces broken default patches for `if`, `contains` and `not`. This isn't addressed there.

And yes, that PR gives me a way to detect the "trivial false schema" cases from an error handler, so I've dropped commit 3.

pjungkamp:
I've added descriptive test cases for both remaining commits. These shouldn't overlap with #384 in intention anymore.

I'd be happy to help get #384 merged (either before or after this PR), but I've got some problems with the approach there and left some suggestions.

pjungkamp:
And please consider merging #370! The other default value processing bugs are mostly niche, but I can't trust the default patch for any schema that contains `"default": null` without that PR...

Issue #352, opened 2025-04-03
Get more actionable error information


persona94:
I overloaded the error handler much like the example here

```
class custom_error_handler : public nlohmann::json_schema::basic_error_handler
    {
        void error(const nlohmann::json_pointer<nlohmann::basic_json<>> &pointer, const json &instance,
            const std::string &message) override
        {
            nlohmann::json_schema::basic_error_handler::error(pointer, instance, message);
            std::cerr << "ERROR: '" << pointer << "' - '" << instance << "': " << message << "\n";
            errorVector.push_back(message);


        }

       std::vector<std::string> errorVector;

    };
 ```

I push the message to a vector inside custom_error_handler and can extract some info from the message by string processing. But given that there can be several kinds of errors, it's not feasible to string process `message` for each of them. 

I understand this lib is for debugging only, but is there anyway to provide an "error type" (perhaps as an enum) in the error? That way the calling function can know what the error was (required missing, min length. max length etc.) and we can extract the property from the error message if needed (although it would be super cool to just get that as well0. I think that would take this lib to the next level.

Sample code

```
    auto path = ptr.to_string();
    std::string property;

    if (message.find("required property") != std::string::npos)
    {
        // For missing required properties, extract property name from the error message
        // Format: "required property 'propertyName' not found"
        size_t start = message.find("'") + 1;
        size_t end = message.find("'", start);
        if (start != std::string::npos && end != std::string::npos)
        {
            property = message.substr(start, end - start);
        }
        
    }
    else
    {
        // Extract the property name from the path for other errors
        std::string path = ptr.to_string();
        size_t last_slash = path.find_last_of('/');
        if (last_slash != std::string::npos)
        {
            property = path.substr(last_slash + 1);
        }
        else if (!path.empty())
        {
            property = path;
        }
        else
        {
            property = "";
        }
    }
```

This code can extract the missing property, or a property that failed the schema due to whatever reason. But in the latter case it's nearly impossible to know what the failure was unless I reverse engineer all the possible error strings that the validation can generate, then handle each of them. Getting the "property" that failed, and a "failure reason" will help the caller of the validation immediately know where the json is wrong. This will allow the caller to perform more granular error handling.

PR #322, opened 2024-05-13
Added expected values to enum and const error messages

bcfjohns:

Issue #321, opened 2024-05-13
Add expected values to const and enum schema failures


bcfjohns:
When a const or enum check fails the message does not include the expected const value, or list of possible enum values. Adding this would make for better error messages.

PR #384, opened 2026-08-27
Make validation errors machine-readable

hughcars:
Addresses #352.

`error_handler` currently exposes an instance path, instance value, and human-readable message. Consumers that need to distinguish failure types or recover constraint data must parse implementation-specific strings.

This adds a structured callback while retaining the existing callback for source compatibility:

```cpp
struct validation_error
{
    json::json_pointer instance_location;
    const json &instance;
    std::string message;
    std::string keyword;
    json details;
};

class error_handler
{
public:
    virtual ~error_handler() {}
    virtual void error(const validation_error &error);
    virtual void error(const json::json_pointer &ptr,
                       const json &instance,
                       const std::string &message);
};
```

The structured callback forwards to the original three-argument callback by default. Existing handlers continue to compile unchanged; new handlers override `error(const validation_error &)`.

`keyword` identifies the failed JSON Schema keyword. `details` carries its parsed constraint and any occurrence-specific context. `instance` is a non-owning reference valid during validation; handlers copy it when retaining error data.

For example, given:

```cpp
const json schema = {
    {"properties",
     {{"fruit",
       {{"enum", {"Apple", "Orange"}}}}}}
};

const json instance = {
    {"fruit", "Potato"}
};
```

a handler can report the available values without parsing the message:

```cpp
class enum_error_handler : public nlohmann::json_schema::error_handler
{
    void error(const nlohmann::json_schema::validation_error &error) override
    {
        if (error.keyword == "enum") {
            std::cerr << "Invalid value " << error.instance
                      << " at " << error.instance_location
                      << "; expected one of "
                      << error.details.at("value") << '\n';
            return;
        }

        std::cerr << error.message << '\n';
    }
};

json_validator validator(schema);
enum_error_handler errors;
validator.validate(instance, errors);
```

This produces:

```text
Invalid value "Potato" at /fruit; expected one of ["Apple","Orange"]
```

Structured details are supplied for scalar, object, array, reference, setup, and logical-combination failures. They survive nested `allOf`, `anyOf`, and `oneOf` propagation. Errors which do not correspond to a schema keyword use an empty keyword and a stable `details["code"]`.

`basic_error_handler` remains the concrete implementation for callers that only need boolean failure state and `reset()`.

The common detail fields are documented in the README. The review surface is concentrated in three files:

- `src/nlohmann/json-schema.hpp` defines the callback contract;
- `src/json-validator.cpp` attaches details at each existing error site;
- `test/issue-352-actionable-errors.cpp` covers each validation family and nested propagation.

The commits are grouped by contract and handler migration, structured behavior, focused regressions, and the compatibility layer requested during review.

## Compatibility

Custom handlers using the existing three-argument callback remain source-compatible. New handlers can override the structured callback directly. The legacy path has direct regression coverage, including handlers derived from `basic_error_handler`.

## Related changes

Expected `enum` and `const` values are now available to custom handlers, related to #321 and #322, without changing their human-readable messages.

The conversion also exposed two existing diagnostics bugs:

- tuple `items` and `additionalItems` failures always reported index zero;
- simultaneous numeric failures were concatenated into one message without a separator.

Both have direct regression coverage. They are included because correct locations and one-keyword-per-error behavior are required by the structured result. I can split them if that would make the change easier to review.

This overlaps parts of #383, particularly internal error collection and object/array validation. I am happy to rebase and reconcile the propagation behavior whichever change lands first.

### DIFF
--- a/src/json-validator.cpp
+++ b/src/json-validator.cpp
@@ -86,7 +86,7 @@ class schema_ref : public schema
 		if (target)
 			target->validate(ptr, instance, patch, e);
 		else
-			e.error(validation_error{ptr, instance, "unresolved or freed schema-reference " + id_, {}, json::object()});
+			e.error(validation_error{ptr, instance, "unresolved or freed schema-reference " + id_, "$ref", {{"value", id_}}});
 	}
 
 	const json &default_value(const json::json_pointer &ptr, const json &instance, error_handler &e) const override final
@@ -98,7 +98,7 @@ class schema_ref : public schema
 		if (target)
 			return target->default_value(ptr, instance, e);
 
-		e.error(validation_error{ptr, instance, "unresolved or freed schema-reference " + id_, {}, json::object()});
+		e.error(validation_error{ptr, instance, "unresolved or freed schema-reference " + id_, "$ref", {{"value", id_}}});
 
 		return default_value_;
 	}
@@ -344,20 +344,20 @@ public:
 	              const json_uri &initial) const
 	{
 		if (!root_) {
-			e.error(validation_error{ptr, "", "no root schema has yet been set for validating an instance", {}, json::object()});
+			e.error(validation_error{ptr, "", "no root schema has yet been set for validating an instance", "", {{"code", "no-root-schema"}}});
 			return;
 		}
 
 		auto file_entry = files_.find(initial.location());
 		if (file_entry == files_.end()) {
-			e.error(validation_error{ptr, "", "no file found serving requested root-URI. " + initial.location(), {}, json::object()});
+			e.error(validation_error{ptr, "", "no file found serving requested root-URI. " + initial.location(), "", {{"code", "schema-file-not-found"}, {"location", initial.location()}}});
 			return;
 		}
 
 		auto &file = file_entry->second;
 		auto sch = file.schemas.find(initial.fragment());
 		if (sch == file.schemas.end()) {
-			e.error(validation_error{ptr, "", "no schema find for request initial URI: " + initial.to_string(), {}, json::object()});
+			e.error(validation_error{ptr, "", "no schema find for request initial URI: " + initial.to_string(), "", {{"code", "schema-not-found"}, {"location", initial.location()}, {"fragment", initial.fragment()}}});
 			return;
 		}
 
@@ -388,6 +388,32 @@ public:
 	operator bool() const { return has_error_; }
 };
 
+class annotating_error_handler : public error_handler
+{
+	error_handler &handler_;
+	std::string keyword_;
+	json keyword_value_;
+	json details_;
+
+public:
+	annotating_error_handler(error_handler &handler, std::string keyword, json keyword_value, json details)
+	    : handler_(handler), keyword_(std::move(keyword)), keyword_value_(std::move(keyword_value)), details_(std::move(details)) {}
+
+	void error(const validation_error &source_error) override
+	{
+		validation_error error = source_error;
+		for (const auto &detail : details_.items())
+			if (!error.details.contains(detail.key()))
+				error.details[detail.key()] = detail.value();
+		if (error.keyword.empty()) {
+			error.keyword = keyword_;
+			if (!keyword_value_.is_null())
+				error.details["value"] = keyword_value_;
+		}
+		handler_.error(error);
+	}
+};
+
 class logical_not : public schema
 {
 	std::shared_ptr<schema> subschema_;
@@ -398,7 +424,7 @@ class logical_not : public schema
 		subschema_->validate(ptr, instance, patch, esub);
 
 		if (!esub)
-			e.error(validation_error{ptr, instance, "the subschema has succeeded, but it is required to not validate", {}, json::object()});
+			e.error(validation_error{ptr, instance, "the subschema has succeeded, but it is required to not validate", "not", json::object()});
 	}
 
 	const json &default_value(const json::json_pointer &ptr, const json &instance, error_handler &e) const override
@@ -471,7 +497,7 @@ class logical_combination : public schema
 		}
 
 		if (count == 0) {
-			e.error(validation_error{ptr, instance, "no subschema has succeeded, but one of them is required to validate. Type: " + key + ", number of failed subschemas: " + std::to_string(subschemata_.size()), {}, json::object()});
+			e.error(validation_error{ptr, instance, "no subschema has succeeded, but one of them is required to validate. Type: " + key + ", number of failed subschemas: " + std::to_string(subschemata_.size()), key, {{"failed_subschemas", subschemata_.size()}}});
 			error_summary.propagate(e, "[combination: " + key + " / ");
 		}
 	}
@@ -507,7 +533,7 @@ bool logical_combination<allOf>::is_validate_complete(const json &, const json::
 {
 	if (esub) {
 		const validation_error &first_error = esub.error_entry_list_.front();
-		e.error(validation_error{first_error.instance_location, first_error.instance, "at least one subschema has failed, but all of them are required to validate - " + first_error.message, {}, json::object()});
+		e.error(validation_error{first_error.instance_location, first_error.instance, "at least one subschema has failed, but all of them are required to validate - " + first_error.message, "allOf", {{"failed_subschema", current_schema_index}}});
 		esub.propagate(e, "[combination: allOf / case#" + std::to_string(current_schema_index) + "] ");
 	}
 	return esub;
@@ -523,13 +549,14 @@ template <>
 bool logical_combination<oneOf>::is_validate_complete(const json &instance, const json::json_pointer &ptr, error_handler &e, const logical_combination_error_handler &, size_t count, size_t)
 {
 	if (count > 1)
-		e.error(validation_error{ptr, instance, "more than one subschema has succeeded, but exactly one of them is required to validate", {}, json::object()});
+		e.error(validation_error{ptr, instance, "more than one subschema has succeeded, but exactly one of them is required to validate", "oneOf", {{"successful_subschemas", count}}});
 	return count > 1;
 }
 
 class type_schema : public schema
 {
 	std::vector<std::shared_ptr<schema>> type_;
+	json typeKeyword_;
 	std::pair<bool, json> enum_, const_;
 	std::vector<std::shared_ptr<schema>> logic_;
 
@@ -548,8 +575,12 @@ class type_schema : public schema
 
 		if (type)
 			type->validate(ptr, instance, patch, e);
-		else
-			e.error(validation_error{ptr, instance, "unexpected instance type", {}, json::object()});
+		else {
+			json details = {{"actual_type", instance.type_name()}};
+			if (!typeKeyword_.is_null())
+				details["value"] = typeKeyword_;
+			e.error(validation_error{ptr, instance, "unexpected instance type", "type", details});
+		}
 
 		if (enum_.first) {
 			bool seen_in_enum = false;
@@ -619,6 +650,8 @@ public:
 		std::set<std::string> known_keywords;
 
 		auto attr = sch.find("type");
+		if (attr != sch.end())
+			typeKeyword_ = attr.value();
 		if (attr == sch.end()) // no type field means all sub-types possible
 			for (auto &t : schema_types)
 				type_[static_cast<uint8_t>(t.second)] = type_schema::make(sch, t.second, root, uris, known_keywords);
@@ -758,7 +791,7 @@ class string : public schema
 			if (utf8_length(instance.get<std::string>()) < minLength_.second) {
 				std::ostringstream s;
 				s << "instance is too short as per minLength:" << minLength_.second;
-				e.error(validation_error{ptr, instance, s.str(), {}, json::object()});
+				e.error(validation_error{ptr, instance, s.str(), "minLength", {{"value", minLength_.second}}});
 			}
 		}
 
@@ -766,22 +799,29 @@ class string : public schema
 			if (utf8_length(instance.get<std::string>()) > maxLength_.second) {
 				std::ostringstream s;
 				s << "instance is too long as per maxLength: " << maxLength_.second;
-				e.error(validation_error{ptr, instance, s.str(), {}, json::object()});
+				e.error(validation_error{ptr, instance, s.str(), "maxLength", {{"value", maxLength_.second}}});
 			}
 		}
 
 		if (std::get<0>(content_)) {
+			const bool has_encoding = !std::get<1>(content_).empty();
+			const std::string keyword = has_encoding ? "contentEncoding" : "contentMediaType";
+			json details = {{"value", has_encoding ? std::get<1>(content_) : std::get<2>(content_)}};
+			if (has_encoding && !std::get<2>(content_).empty())
+				details["content_media_type"] = std::get<2>(content_);
+
 			if (root_->content_check() == nullptr)
-				e.error(validation_error{ptr, instance, std::string("a content checker was not provided but a contentEncoding or contentMediaType for this string have been present: '") + std::get<1>(content_) + "' '" + std::get<2>(content_) + "'", {}, json::object()});
+				e.error(validation_error{ptr, instance, std::string("a content checker was not provided but a contentEncoding or contentMediaType for this string have been present: '") + std::get<1>(content_) + "' '" + std::get<2>(content_) + "'", keyword, details});
 			else {
 				try {
 					root_->content_check()(std::get<1>(content_), std::get<2>(content_), instance);
 				} catch (const std::exception &ex) {
-					e.error(validation_error{ptr, instance, std::string("content-checking failed: ") + ex.what(), {}, json::object()});
+					details["reason"] = ex.what();
+					e.error(validation_error{ptr, instance, std::string("content-checking failed: ") + ex.what(), keyword, details});
 				}
 			}
 		} else if (instance.type() == json::value_t::binary) {
-			e.error(validation_error{ptr, instance, "expected string, but get binary data", {}, json::object()});
+			e.error(validation_error{ptr, instance, "expected string, but get binary data", "type", {{"value", "string"}, {"actual_type", "binary"}}});
 		}
 
 		if (instance.type() != json::value_t::string) {
@@ -790,18 +830,21 @@ class string : public schema
 
 #ifndef NO_STD_REGEX
 		if (pattern_.first &&
-		    !REGEX_NAMESPACE::regex_search(instance.get<std::string>(), pattern_.second))
-			e.error(validation_error{ptr, instance, "instance does not match regex pattern: " + patternString_, {}, json::object()});
+		    !REGEX_NAMESPACE::regex_search(instance.get<std::string>(), pattern_.second)) {
+			json details = json::object();
+			details["value"] = patternString_;
+			e.error(validation_error{ptr, instance, "instance does not match regex pattern: " + patternString_, "pattern", details});
+		}
 #endif
 
 		if (format_.first) {
 			if (root_->format_check() == nullptr)
-				e.error(validation_error{ptr, instance, std::string("a format checker was not provided but a format keyword for this string is present: ") + format_.second, {}, json::object()});
+				e.error(validation_error{ptr, instance, std::string("a format checker was not provided but a format keyword for this string is present: ") + format_.second, "format", {{"value", format_.second}}});
 			else {
 				try {
 					root_->format_check()(format_.second, instance.get<std::string>());
 				} catch (const std::exception &ex) {
-					e.error(validation_error{ptr, instance, std::string("format-checking failed: ") + ex.what(), {}, json::object()});
+					e.error(validation_error{ptr, instance, std::string("format-checking failed: ") + ex.what(), "format", {{"value", format_.second}, {"reason", ex.what()}}});
 				}
 			}
 		}
@@ -885,6 +928,7 @@ class numeric : public schema
 	bool exclusiveMinimum_ = false;
 
 	std::pair<bool, json::number_float_t> multipleOf_{false, 0};
+	json multipleOfKeyword_;
 
 	// multipleOf - if the remainder of the division is 0 -> OK
 	bool violates_multiple_of(T x) const
@@ -903,31 +947,34 @@ class numeric : public schema
 	{
 		T value = instance; // conversion of json to value_type
 
-		std::ostringstream oss;
-
-		if (multipleOf_.first && value != 0) // zero is multiple of everything
-			if (violates_multiple_of(value))
-				oss << "instance is not a multiple of " << json(multipleOf_.second);
+		if (multipleOf_.first && value != 0 && violates_multiple_of(value)) { // zero is multiple of everything
+			std::ostringstream message;
+			message << "instance is not a multiple of " << json(multipleOf_.second);
+			e.error(validation_error{ptr, instance, message.str(), "multipleOf", {{"value", multipleOfKeyword_}}});
+		}
 
 		if (maximum_.first) {
-			if (exclusiveMaximum_ && value >= maximum_.second)
-				oss << "instance exceeds or equals maximum of " << json(maximum_.second);
-			else if (value > maximum_.second)
-				oss << "instance exceeds maximum of " << json(maximum_.second);
+			if (exclusiveMaximum_ && value >= maximum_.second) {
+				std::ostringstream message;
+				message << "instance exceeds or equals maximum of " << json(maximum_.second);
+				e.error(validation_error{ptr, instance, message.str(), "exclusiveMaximum", {{"value", maximum_.second}}});
+			} else if (value > maximum_.second) {
+				std::ostringstream message;
+				message << "instance exceeds maximum of " << json(maximum_.second);
+				e.error(validation_error{ptr, instance, message.str(), "maximum", {{"value", maximum_.second}}});
+			}
 		}
 
 		if (minimum_.first) {
-			if (exclusiveMinimum_ && value <= minimum_.second)
-				oss << "instance is below or equals minimum of " << json(minimum_.second);
-			else if (value < minimum_.second)
-				oss << "instance is below minimum of " << json(minimum_.second);
-		}
-
-		oss.seekp(0, std::ios::end);
-		auto size = oss.tellp();
-		if (size != 0) {
-			oss.seekp(0, std::ios::beg);
-			e.error(validation_error{ptr, instance, oss.str(), {}, json::object()});
+			if (exclusiveMinimum_ && value <= minimum_.second) {
+				std::ostringstream message;
+				message << "instance is below or equals minimum of " << json(minimum_.second);
+				e.error(validation_error{ptr, instance, message.str(), "exclusiveMinimum", {{"value", minimum_.second}}});
+			} else if (value < minimum_.second) {
+				std::ostringstream message;
+				message << "instance is below minimum of " << json(minimum_.second);
+				e.error(validation_error{ptr, instance, message.str(), "minimum", {{"value", minimum_.second}}});
+			}
 		}
 	}
 
@@ -964,6 +1011,7 @@ public:
 		attr = sch.find("multipleOf");
 		if (attr != sch.end()) {
 			multipleOf_ = {true, attr.value().get<json::number_float_t>()};
+			multipleOfKeyword_ = attr.value();
 			kw.insert("multipleOf");
 		}
 	}
@@ -974,7 +1022,7 @@ class null : public schema
 	void validate(const json::json_pointer &ptr, const json &instance, json_patch &, error_handler &e) const override
 	{
 		if (!instance.is_null())
-			e.error(validation_error{ptr, instance, "expected to be null", {}, json::object()});
+			e.error(validation_error{ptr, instance, "expected to be null", "type", {{"value", "null"}, {"actual_type", instance.type_name()}}});
 	}
 
 public:
@@ -997,15 +1045,7 @@ class boolean : public schema
 	void validate(const json::json_pointer &ptr, const json &instance, json_patch &, error_handler &e) const override
 	{
 		if (!true_) { // false schema
-			// empty array
-			// switch (instance.type()) {
-			// case json::value_t::array:
-			//	if (instance.size() != 0) // valid false-schema
-			//		e.error(validation_error{ptr, instance, "false-schema required empty array", {}, json::object()});
-			//	return;
-			//}
-
-			e.error(validation_error{ptr, instance, "instance invalid as per false-schema", {}, json::object()});
+			e.error(validation_error{ptr, instance, "instance invalid as per false-schema", "", {{"code", "false-schema"}}});
 		}
 	}
 
@@ -1016,18 +1056,19 @@ public:
 
 class required : public schema
 {
+	const std::string property_;
 	const std::vector<std::string> required_;
 
 	void validate(const json::json_pointer &ptr, const json &instance, json_patch &, error_handler &e) const override final
 	{
 		for (auto &r : required_)
 			if (instance.find(r) == instance.end())
-				e.error(validation_error{ptr, instance, "required property '" + r + "' not found in object as a dependency", {}, json::object()});
+				e.error(validation_error{ptr, instance, "required property '" + r + "' not found in object as a dependency", "dependencies", {{"value", required_}, {"property", property_}, {"missing_property", r}}});
 	}
 
 public:
-	required(const std::vector<std::string> &r, root_schema *root)
-	    : schema(root), required_(r) {}
+	required(const std::string &property, const std::vector<std::string> &r, root_schema *root)
+	    : schema(root), property_(property), required_(r) {}
 };
 
 class object : public schema
@@ -1041,27 +1082,31 @@ class object : public schema
 	std::vector<std::pair<REGEX_NAMESPACE::regex, std::shared_ptr<schema>>> patternProperties_;
 #endif
 	std::shared_ptr<schema> additionalProperties_;
+	json additionalPropertiesKeyword_;
 
 	std::map<std::string, std::shared_ptr<schema>> dependencies_;
 
 	std::shared_ptr<schema> propertyNames_;
+	json propertyNamesKeyword_;
 
 	void validate(const json::json_pointer &ptr, const json &instance, json_patch &patch, error_handler &e) const override
 	{
 		if (maxProperties_.first && instance.size() > maxProperties_.second)
-			e.error(validation_error{ptr, instance, "too many properties", {}, json::object()});
+			e.error(validation_error{ptr, instance, "too many properties", "maxProperties", {{"value", maxProperties_.second}}});
 
 		if (minProperties_.first && instance.size() < minProperties_.second)
-			e.error(validation_error{ptr, instance, "too few properties", {}, json::object()});
+			e.error(validation_error{ptr, instance, "too few properties", "minProperties", {{"value", minProperties_.second}}});
 
 		for (auto &r : required_)
 			if (instance.find(r) == instance.end())
-				e.error(validation_error{ptr, instance, "required property '" + r + "' not found in object", {}, json::object()});
+				e.error(validation_error{ptr, instance, "required property '" + r + "' not found in object", "required", {{"value", required_}, {"missing_property", r}}});
 
 		// for each property in instance
 		for (auto &p : instance.items()) {
-			if (propertyNames_)
-				propertyNames_->validate(ptr, p.key(), patch, e);
+			if (propertyNames_) {
+				annotating_error_handler property_name_error(e, "propertyNames", propertyNamesKeyword_, {{"property", p.key()}});
+				propertyNames_->validate(ptr, p.key(), patch, property_name_error);
+			}
 
 			bool a_prop_or_pattern_matched = false;
 			auto schema_p = properties_.find(p.key());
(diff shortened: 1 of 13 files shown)


pboettch:
> The existing location, instance, and message remain available on the error object. If preserving source compatibility with the current callback is preferred, I can adapt this to provide a forwarding path.

Yes please. Keep this small layer of compat. It is easier to maintain this than communicating on breaking changes.

pboettch:
I like your approach and your reasoning.

pboettch (approved):

pboettch:
The structure part will allow additional changes in the future without breaking compat, nice.

hughcars:
Done. The structured callback now forwards to the original three-argument callback by default, so existing handlers continue to compile unchanged. I added a regression using the legacy override through `basic_error_handler`; both nlohmann/json 3.8.0 and 3.12.0 pass locally.

I also updated the README and PR description to describe the compatibility path. Would you like the old path marked deprecated to encourage migrating it?

pboettch:
OK, what's missing to undraft it?

src/json-validator.cpp
```diff
@@ -1161,6 +1212,8 @@ class object : public schema
 
 		attr = sch.find("additionalProperties");
 		if (attr != sch.end()) {
+			if (attr.value().is_boolean())
+				additionalPropertiesKeyword_ = attr.value();
 			additionalProperties_ = schema::make(attr.value(), root, {"additionalProperties"}, uris);
```
pjungkamp: This means that `additionalPropertiesKeyword_` can only indicate three distinct cases:

- `additionalPropertiesKeyword_.is_null()` means that `additionalProperties_` is some arbtrary schema.
- `additionalPropertiesKeyword_ == true` means that `additionalProperties_` is a `boolean_schema` with value `true`, will never fail, and is thus completely redundant.
- `additionalPropertiesKeyword_ == false` means that `additionalProperties_` is a `boolean_schema` with value `false` and will always fail.

I'd propose to store these three cases a little differently. Just add a `bool noAdditionalProperties_ = false;` member and do the assignment like this:

```suggestion
			auto additionalProperties = schema::make(attr.value(), root, {"additionalProperties"}, uris);
			if (attr.value().is_boolean())
				denyAdditionalProperties_ = !attr.value();
			else
				additionalProperties_ = additionalProperties;
```

This would only need some special handling for `denyAdditionalProperties_` as I described above.

src/json-validator.cpp
```diff
@@ -1047,27 +1082,31 @@ class object : public schema
 	std::vector<std::pair<REGEX_NAMESPACE::regex, std::shared_ptr<schema>>> patternProperties_;
 #endif
 	std::shared_ptr<schema> additionalProperties_;
+	json additionalPropertiesKeyword_;
```
pjungkamp:
```suggestion
	bool denyAdditionalProperties_{false};
```

See my explanation below for why this is sufficient to represent all cases.

src/json-validator.cpp
```diff
@@ -1090,8 +1129,20 @@ class object : public schema
...
+					if (!error.details.contains("property"))
+						error.details["property"] = p.key();
+					if (error.keyword.empty()) {
+						error.keyword = "additionalProperties";
+						if (!additionalPropertiesKeyword_.is_null())
+							error.details["value"] = additionalPropertiesKeyword_;
+					}
+					e.error(error);
+				}
 			}
```
pjungkamp: I'd propose something like this instead:

```suggestion
			if (!a_prop_or_pattern_matched) {
				if (additionalProperties_) {
					annotating_error_handler additional_properties_error(e, "additionalProperties", nullptr, json::object());;
					additionalProperties_->validate(ptr / p.key(), p.value(), patch, additional_properties_error);
				} else if (denyAdditionalProperties_) {
					validation_error error{ptr,	instance, "unexpected additional property '" + p.key() + "'", {{"property", p.key()}}};
					e.error(error);
				}
			}
```

See my comment below for the rationale here.

pjungkamp (commented):
I've got some problems with the approach taken here:

### `*Keyword` members

The code adds multiple `*Keyword_` member variables to the validator objects (e.g. `additionalPropertiesKeyword_`/`propertyNamesKeyword_`/etc.) but these will only ever be `true`, `false` or `null`.

Note that `true` will by definition *never fail* so storing it is redundant. So the only remaining cases are `null` and `false`. I'd propose that you store something like `bool deny*_` (e.g. `denyAdditionalProperties_`/`denyPropertyNames_`) instead.

I've left some code suggestions showing how I'd want this to work for e.g. `additionalProperties`.

### `additionalItems` vs. `additionalProperties` in `details`

Why does `additionalProperties` add a `property` member into details while `additionalItems` does not add an `index` member? This seems inconsistent. I think that you should remove the "property" member from the details unless the `additionalProperties` schema is just `false`. See my suggestions for the additionalProperties handling.

---

**NOTE:** I'm not responsible for this project, don't apply my suggestions, unless @pboettch agreed first. I've tried to improve the handling of some of the `false`-schema cases in #383 myself, namely for `additionalItems` and `additionalProperties`, but you've got the chance here to address _all_ the boolean schema validation edge cases at the same time in all those places.

pjungkamp:
I don't quite like the `keyword` member of the `validation_error` struct. I think I'd prefer something like a pseudo-JSON-path `schema_location` where each segment represents a keyword, index or property name that has been traversed during validation. I'm mentioning this now because all changes that modify or remove any existing member of `struct validation_error` would probably be considered breaking changes.

---

Consider something like the following schema:

```
{
  "type": "object",
  "properties": {
    "foo": {
      "type": array,
      "items": [
        {
          "type": "integer",
          "exclusiveMinimum": 0
        }
      ]
    }
  },
  "additionalItems": {
    "type": array,
    "items": [],
    "additionalItems": {
      "type": "integer",
      "const": 0
    }
  }
}
```

Let's validate `{ "foo": [0], "bar": [1] }`. This should give us two validation errors:

- The first one for the `schema_location` at `/properties/foo/items/0/exclusiveMinimum` and an `instance_location` of `/foo/0`.
- The second one for the `schema_location` at `/additionalProperties/additionalItems/const` and an `instance_location` of `/bar/0`.

This lets me provide targeted suggestions for different parts of my schema. e.g. providing a suggestion that only applies to the keyword at `/properties/foo/items/0/exclusiveMinimum` but not to any other `exclusiveMinimum` keyword in schema.

This also extends nicely to a `false` root schema where the `schema_location` or your `std::string keyword` would just be empty.

hughcars:
@pjungkamp Thanks for taking a close look at this, especially given the overlap with #383. I’ve replied to the implementation points inline.

The schema-location question is the broader one: I think it complements `keyword` rather than replacing it. `keyword` lets a handler dispatch directly:

```cpp
if (error.keyword == "enum")
    report_choices(error.details.at("value"));
```

A schema location answers a separate question: which occurrence of that keyword failed.

I would want to define that carefully around `$id`, `$ref`, and external schemas. The newer JSON Schema output model distinguishes a relative validation path from an absolute resolved schema location ([2020-12 §12](https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-01#section-12)). This validator also carries multiple `json_uri`s while constructing and resolving schemas ([`schema::make`](https://github.com/pboettch/json-schema-validator/blob/c97338b529fcf932571b093429af3d7b34ccc0ff/src/json-validator.cpp#L1420-L1455)), so a root-relative pointer alone will not cover every case.

My preference is to keep `keyword` here and design schema location separately against those semantics. If it belongs in the initial public shape, I can add it once we agree whether that means the evaluation path, resolved URI, or both. Where we disagree I’d defer to @pboettch’s position, I’m just happy to contribute to what is proving a very useful project for us over at Palace.

src/json-validator.cpp
hughcars: You’re right that the retained value is only useful when the Boolean schema is `false`; `true` cannot produce an error. A `bool denyAdditionalProperties_` is clearer than a tri-state `json`.

I would still keep validation going through `schema::make` and `boolean_schema`. That path registers the subschema URI and owns the existing false-schema message and code ([implementation](https://github.com/pboettch/json-schema-validator/blob/c97338b529fcf932571b093429af3d7b34ccc0ff/src/json-validator.cpp#L1042-L1048)).

The smaller change would be:

```cpp
denyAdditionalProperties_ =
    attr.value().is_boolean() && !attr.value().get<bool>();
additionalProperties_ =
    schema::make(attr.value(), root, {"additionalProperties"}, uris);
```

The flag would only control whether `details["value"] = false` is attached. I can apply the same cleanup to `propertyNames` and `additionalItems` if @pboettch wants it here.

src/json-validator.cpp
hughcars: The difference comes from the existing location behavior.

`additionalItems` validates at `ptr / index`, so the error already reports `/1` ([code](https://github.com/pboettch/json-schema-validator/blob/c97338b529fcf932571b093429af3d7b34ccc0ff/src/json-validator.cpp#L1290-L1311), [test](https://github.com/pboettch/json-schema-validator/blob/c97338b529fcf932571b093429af3d7b34ccc0ff/test/issue-352-actionable-errors.cpp#L133-L147)). Adding an `index` would duplicate that.

`additionalProperties` validates at `ptr / property`, but its wrapper then resets the location to the containing object to preserve existing behavior ([code](https://github.com/pboettch/json-schema-validator/blob/c97338b529fcf932571b093429af3d7b34ccc0ff/src/json-validator.cpp#L1128-L1145)). Without `details["property"]`, the offending property would be lost.

Preserving `/extra` instead would be cleaner, and then `property` may be unnecessary for nested-schema failures. That is a separate observable behavior change though, I would not remove `property` without making that location change at the same time.

src/json-validator.cpp
hughcars: Agreed on the member representation. I would use `denyAdditionalProperties_` as the marker, but keep the schema pointer and generic validation path; I outlined the smaller version [here](https://github.com/pboettch/json-schema-validator/pull/384#discussion_r3881239654).

src/json-validator.cpp
pjungkamp: That location change was what I wanted to introduce with #383. I think it's inconsistent to do the location reset for `additionalProperties` but not for `additionalItems`.

This behavior is especially problematic because the `additonalProperties` validation through the `first_error_handler` is the only validation path that discards unexpected errors without forwarding them to the user's error handler.

And yes, that does affect observable behavior, I had to apply changes to two regression test while implementing #383.

src/json-validator.cpp
pjungkamp: We're checking for a boolean here anyway to determine the value of `additonalProperties_` shouldn't we take the opportunity to filter out the `true` schema case?

We're linting our schemas to require an explicit `additionalProperties` keyword next to every `properties` keyword, even when the value is `true` to indicate to the reader that allowing additional properties was a conscious decision. So I'd like to see some trivial filtering for these cases.

I can also propose that in a separate follow up PR, but I wanted to at least discuss it here first because I'd have to touch every boolean-schema case you're changeing here again.

And on a more personal note: I hate the error messages for `"additionalProperties": false`., e.g. _"validation failed for additonal property "XXX": instance invalid as per false-schema"_, and I really want something like _"unexpected additional property"_ instead.

pjungkamp:
The keyword location in the newer standard's revisions was was inspired me. But your reasoning on keeping the simple keyword for now seems reasonable.

But then I've got another question: Should `keyword` be a non-static data member or should it be part of the `details`? There is a valid schema where this metadata could be absent/empty (when the root schema is `false`). 

And let's assume that we'd add a keyword path to the `details` at some point. The keyword would by definiton just be the last element of that path, but we have to keep both for backwards compatibility. Just some some food for thought here...

src/json-validator.cpp
pjungkamp: Also note that `if (!error.details.contains("property")) error.details["property"] = p.key();` and the `if (error.keyword.empty()) error.keyword = "additionalProperties";` as currently implemented is almost certainly undesirable behavior. Let's assume the following schema:

```
{
 "type": "object",
 "additionalProperties": {
  "type": "object",
  "propertyNames": false
  } 
}
```

And let validate this value: `{ "foo": { "bar": "baz" }}`

The `"bar"` property will fail the validation for the `"propertyNames"` keyword and then propagate the error through your logic for the additional properties error handling. The resulting error for what you've currently implemented:

```cpp
validation_error {
  .instance_location = "/",
  .instance = {"foo", {{"bar", "baz"}}},
  .keyword = "propertyNames",
  .details = {{"value": false}, {"property", "bar"}},
}
``` 

The details and keyword tell us that there was a problem with the property name `"bar"` but there isn't a `"bar"` property in the specified instance. You can either discard the error details and keyword from the nested error, or you apply my change where you properly report all errors for additional properties... 

_The current error reporting for additional properties is just bad._

src/json-validator.cpp
```diff
@@ -402,7 +424,7 @@ class logical_not : public schema
 		subschema_->validate(ptr, instance, patch, esub);
 
 		if (!esub)
-			e.error(ptr, instance, "the subschema has succeeded, but it is required to not validate");
+			e.error(validation_error{ptr, instance, "the subschema has succeeded, but it is required to not validate", "not", json::object()});
```
pjungkamp, line 403: Should this error also include `value` in details for `"not": true`?

src/json-validator.cpp
```diff
@@ -513,7 +532,8 @@ template <>
 bool logical_combination<allOf>::is_validate_complete(const json &, const json::json_pointer &, error_handler &e, const logical_combination_error_handler &esub, size_t, size_t current_schema_index)
 {
 	if (esub) {
-		e.error(esub.error_entry_list_.front().ptr_, esub.error_entry_list_.front().instance_, "at least one subschema has failed, but all of them are required to validate - " + esub.error_entry_list_.front().message_);
+		const validation_error &first_error = esub.error_entry_list_.front();
+		e.error(validation_error{first_error.instance_location, first_error.instance, "at least one subschema has failed, but all of them are required to validate - " + first_error.message, "allOf", {{"failed_subschema", current_schema_index}}});
```
pjungkamp, line 512: Should this error also include `value` in details when that specific subschema is `false`?

src/json-validator.cpp
```diff
@@ -529,13 +549,14 @@ template <>
...
 	if (count > 1)
-		e.error(ptr, instance, "more than one subschema has succeeded, but exactly one of them is required to validate");
+		e.error(validation_error{ptr, instance, "more than one subschema has succeeded, but exactly one of them is required to validate", "oneOf", {{"successful_subschemas", count}}});
 	return count > 1;
 }
 
 class type_schema : public schema
 {
 	std::vector<std::shared_ptr<schema>> type_;
+	json typeKeyword_;
```
pjungkamp, line 535: If you choose to adopt `bool deny*_` for `additionalProperties`/etc you may also want to consider whether this value here makes sense.

This value will essentially convey two states: `type` is an array and `type` is a string. The actual array or string value for `type` can easily be reconstructed from the `type_` array.

A string and a single-item array are equivalent from a schema perspective, so I don't know whether a distinction would even be desirable here. Should an error handler ever care whether the schema said `"type": ["string"]` or `"type": "string"`? We could just report an array value unconditionally.

src/json-validator.cpp
```diff
@@ -554,8 +575,12 @@ class type_schema : public schema
 
 		if (type)
 			type->validate(ptr, instance, patch, e);
-		else
-			e.error(ptr, instance, "unexpected instance type");
+		else {
+			json details = {{"actual_type", instance.type_name()}};
+			if (!typeKeyword_.is_null())
```
pjungkamp: The `type` validation can never fail when there is no `type` keyword in the schema. This condition should never be false.

src/json-validator.cpp
```diff
@@ -796,18 +830,21 @@ class string : public schema
 
 #ifndef NO_STD_REGEX
 		if (pattern_.first &&
-		    !REGEX_NAMESPACE::regex_search(instance.get<std::string>(), pattern_.second))
-			e.error(ptr, instance, "instance does not match regex pattern: " + patternString_);
+		    !REGEX_NAMESPACE::regex_search(instance.get<std::string>(), pattern_.second)) {
+			json details = json::object();
+			details["value"] = patternString_;
+			e.error(validation_error{ptr, instance, "instance does not match regex pattern: " + patternString_, "pattern", details});
```
pjungkamp: Why is this called `pattern` and not `value`?

src/json-validator.cpp
```diff
@@ -1260,7 +1323,7 @@ class array : public schema
 				}
 			}
 			if (!contained)
-				e.error(ptr, instance, "array does not contain required element as per 'contains'");
+				e.error(validation_error{ptr, instance, "array does not contain required element as per 'contains'", "contains", json::object()});
```
pjungkamp, line 1290: Should this error also include `value` in details for `"contains": false`?

src/json-validator.cpp
hughcars: Implemented in [`52674ff`](https://github.com/pboettch/json-schema-validator/pull/384/commits/52674ff0a9d18189073a80df974924d113169755): `true` is skipped, `false` uses a `deny*` flag and direct error, and non-Boolean schemas retain the validator. I applied this to all three applicators with tests. See if this matches what you had in mind.

src/json-validator.cpp
hughcars: Implemented in [`422a525`](https://github.com/pboettch/json-schema-validator/pull/384/commits/422a5250b71a71923578abd72cfe318ce272f0bf): nested errors now keep their child location and are all forwarded. I added a regression proving both errors arrive. See if you agree with the implementation.

pjungkamp (commented):
I've taken a closer look at all error `details` and noticed some more discrepancies.

You don't handle _all_ boolean subschemas: `not`, `contains`, `properties`, `dependencies`, `items`, `allOf`, `oneOf` and `anyOf` are currently missing a `value` in details when a `subschema` is a boolean.

src/json-validator.cpp
```diff
@@ -374,22 +374,44 @@ namespace
 class first_error_handler : public error_handler
 {
 public:
-	bool error_{false};
-	json::json_pointer ptr_;
-	json instance_;
-	std::string message_;
+	bool has_error_{false};
+	validation_error first_error_{json::json_pointer{}, nullptr, {}, {}, json::object()};
```
pjungkamp: This field isn't used anymore. The only place that read the error message was the flawed `additionalProperties` validation and you replaced that with an `annotating_error_handler`.

src/json-validator.cpp
pjungkamp: Yes, absolutely! I've just tested it, and I'm rebasing my other changes onto your branch. :D

pjungkamp:
This isn't quite optimal for validating large JSON values. `validation_error` stores `json instance` by value and thus requires a copy of the entire instance for each validation error. And to reiterate: Some validation errors are an expected part of any schema validation using `if`, `not`, `oneOf`, `anyOf` or `contains`, where a subschema might even fail more often than it ever succeeds but still results in the value passing validation.

This is the same problem that `logical_combination_error_handler::error_entry` currently has: Validating a e.g. a `oneOf` with N branches is _guaranteed to create at least N-1 copies_ of some json instance that is stored for later propagation (this is also the case on `main`...).

Could we make the `instance` member of `validation_error` a const reference? The validated instance is assumed to be constant for the duration of the validation anyway, so a copy is completely unnecessary.

<details>
<summary>Patches</summary>

#### Patch for this PR

```patch 
diff --git a/src/nlohmann/json-schema.hpp b/src/nlohmann/json-schema.hpp
index cb1d723..c7ac027 100644
--- a/src/nlohmann/json-schema.hpp
+++ b/src/nlohmann/json-schema.hpp
@@ -139,7 +139,7 @@ typedef std::function<void(const std::string & /*contentEncoding*/, const std::s
 // including the schema keyword value when available, belongs in details.
 struct validation_error {
 	json::json_pointer instance_location;
-	json instance;
+	json const &instance;
 	std::string message;
 	std::string keyword;
 	json details;
```

#### Patch for `logical_combination_error_handler::error_entry` on `main`

```patch
diff --git a/src/json-validator.cpp b/src/json-validator.cpp
index 62ce97c..b8e0485 100644
--- a/src/json-validator.cpp
+++ b/src/json-validator.cpp
@@ -431,7 +431,7 @@ class logical_combination_error_handler : public error_handler
 public:
 	struct error_entry {
 		json::json_pointer ptr_;
-		json instance_;
+		const json &instance_;
 		std::string message_;
 	};
```


</details>

<details>
<summary>A small bpftrace allocation profile of `test/json-schema-validate`</summary>

#### setup

I've compiled the `json-schema-validate` test executable with `CMAKE_BUILD_TYPE=RelWithDebInfo` and then run it like this:

```
sudo bpftrace ./bpfscript -c './build/test/json-schema-validate schema.json' < instance.json
```

#### bpftrace script

```
uprobe:./build/test/json-schema-validate:*json_schema*json_validator*validate*json_uri*
{
    @in_target[tid] = 1;
}

uretprobe:./build/test/json-schema-validate:*json_schema*json_validator*validate*json_uri*
{
    delete(@in_target[tid]);
}

uprobe:/nix/store/ja86dg5ja2fxf18sbz8mwvnxbwambpp0-glibc-x86_64-unknown-linux-gnu-2.42-67/lib/libc.so.6:malloc
/@in_target[tid]/
{
    @target_alloc_sizes = hist(arg0);
    @target_alloc_count = count();
}
```

#### example schema

This mirrors a style of schema that we use to discriminate objects based on a `"type"` property.

```json
{
  "type": "object",
  "additionalProperties": {
    "oneOf": [
      {
        "type": "object",
        "properties": {
          "type": {
            "type": "string",
            "const": "foo"
          },
          "bar": {
            "type": "string",
            "const": "baz"
          }
        },
        "additionalProperties": false
      },
      {
        "type": "object",
        "properties": {
          "type": {
            "type": "string",
            "const": "bar"
          },
          "baz": {
            "type": "string",
            "const": "bar"
          }
        },
        "additionalProperties": false
      },
      {
        "type": "object",
        "properties": {
          "type": {
            "type": "string",
            "const": "baz"
          },
          "foo": {
            "type": "string",
            "const": "baz"
          }
        },
        "additionalProperties": false
      }
    ]
  }
}
```

#### example instance

This instance is supposed to validate successfully. 

```
{
  "a": {
    "type": "foo",
    "bar": "baz"
  },
  "b": {
    "type": "bar",
    "baz": "bar"
  },
  "c": {
    "type": "baz",
    "foo": "baz"
  },
  "d": {
    "type": "foo",
    "bar": "baz"
  },
  "e": {
    "type": "bar",
    "baz": "bar"
  },
  "f": {
    "type": "baz",
    "foo": "baz"
  },
  "g": {
    "type": "foo",
    "bar": "baz"
  }
}
```

#### `main` branch

```
@target_alloc_count: 597
@target_alloc_sizes:
[16, 32)             212 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@   |
[32, 64)             224 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
[64, 128)            133 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@                      |
[128, 256)            21 |@@@@                                                |
[256, 512)             7 |@                                                   |
```

#### `validation_error::instance` by value (this PR)

```
@target_alloc_count: 1003
@target_alloc_sizes:
[16, 32)             506 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
[32, 64)             336 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@                  |
[64, 128)            133 |@@@@@@@@@@@@@                                       |
[128, 256)            21 |@@                                                  |
[256, 512)             7 |                                                    |
```

### `main` branch with patch

```
@target_alloc_count: 401
@target_alloc_sizes:
[16, 32)             100 |@@@@@@@@@@@@@@@@@@@@@@@                             |
[32, 64)             217 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
[64, 128)             77 |@@@@@@@@@@@@@@@@@@                                  |
[128, 256)             7 |@                                                   |
```

#### `validation_error::instance` by reference (this PR with patch)

```
@target_alloc_count: 891
@target_alloc_sizes:
[16, 32)             394 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
[32, 64)             336 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@        |
[64, 128)            133 |@@@@@@@@@@@@@@@@@                                   |
[128, 256)            21 |@@                                                  |
[256, 512)             7 |                                                    |
```

### evaluation

This PR increases the total amount of allocations for a successful validation in this example by about 68% (406 allocations) compared to the `main` branch.

The simple patch that takes the `instance` by reference reduces the count by about 112 allocations for this PR and 196 allocations for the equivalent patch on main. This also means that this branch _with this patch_ has about 122% more allocations than the patched main branch _for a successful validation_.

### optimization potential

I've experimented with some more optimizations and got this branch with the structured error messages down to 611 total allocations, just 14 more than the current `main`.
I did this by optimizing the failure cases to just fail faster. Therefore, I think that the better errors are well worth the little extra allocation overhead. There is still a lot of performance to gain in other places that can easily offset this change.

</details>

---

The alternative would be to move `instance` out of the `validation_error` struct and make it a parameter to `error_handler::error` again.

hughcars:
Agreed, thanks for profiling this. Implemented in [`0aaf10c`](https://github.com/pboettch/json-schema-validator/pull/384/commits/0aaf10c): `validation_error::instance` is now a non-owning `const json &`, matching the old callback lifetime. The README documents that retained handlers must copy it, and the test collector now copies into an explicitly owning record. The full Spack test suite passes against nlohmann/json 3.8.0 and 3.12.0.

hughcars:
@pboettch I think this is ready for another look. The requested compatibility and follow-up review changes are in at `0aaf10c`, with local tests and Fedora builds passing. Let me know if anything else is needed.

pboettch:
I'm heart-broken for several reasons ... I'm out of time and energy, so I asked claude code of what he thinks about the PR's discussion and implementation before merging, he found:

The shape now matches what was discussed, but the last commit introduces a memory-safety bug, and there are two smaller correctness points.

### 1. Use-after-scope when a `propertyNames` error propagates through `allOf` / `anyOf` / `oneOf`

`0aaf10c` changes `validation_error::instance` to `const json &`. For `propertyNames` the "instance" being validated is not part of the user's document: it is a temporary `json` constructed from `p.key()` in `object::validate`, both in the `denyPropertyNames_` branch and in the `propertyNames_->validate(ptr, p.key(), ...)` call.

`logical_combination_error_handler` stores a copy of the `validation_error` in `error_entry_list_`, so it keeps a reference to that temporary. By the time `propagate()` forwards the entry to the user's handler, the temporary is gone. On `main` this was safe because `error_entry` held `json instance_` by value.

Repro (any handler that reads `error.instance`, e.g. the README's `custom_error_handler`):

```json
{ "oneOf": [ { "type": "object", "propertyNames": false }, { "type": "string" } ] }
```

validated against

```json
{ "somekey": 1 }
```

gives

```
==ERROR: AddressSanitizer: stack-use-after-return
    #0 ... nlohmann::detail::serializer<...>::dump(...)
    #2 ... custom_error_handler::error(nlohmann::json_schema::validation_error const&)
    #3 ... logical_combination_error_handler::propagate  src/json-validator.cpp:442
    #4 ... logical_combination<oneOf>::validate          src/json-validator.cpp:477
    #5 ... type_schema::validate                         src/json-validator.cpp:576
Address ... is located in stack of thread T0 ... in frame object::validate src/json-validator.cpp:1066
```

The same happens with a real `propertyNames` subschema (e.g. `{"maxLength": 2}`) through `details_error_handler`. The default throwing handler hides it only because it throws on the summary error before `propagate()` runs.

The existing tests pass cleanly under ASan/UBSan, so the missing case is `propertyNames` nested inside a logical combination. Please add that as a regression.

Possible fixes:

- make `logical_combination_error_handler` own the instance again (a small owning entry type, with the reference only materialised when forwarding to the user handler), or
- validate `propertyNames` against a `json` that outlives the error (e.g. build the key `json` once per property and keep it alive across the `validate` call), and document that `instance` may reference validator-internal storage.

The second option keeps the allocation win from the profiling but is easier to get wrong again later, so I'd lean towards the first.

### 2. `validation_error` is neither copy-assignable nor default-constructible

That is a consequence of the reference member. Inside the validator it is fine (only copy-construction and `push_back` are used), but this is a public struct and users will put it into containers and then `erase`, `sort`, or reassign. The alternative raised during review, passing `instance` as a separate parameter of `error()` and keeping the struct value-only, avoids both this and finding 1. If the reference member stays, please at least call this out in the README next to the "copy it if you retain it" note.

### 3. Undeclared ABI break

A new virtual `error(const validation_error &)` is inserted before the existing three-argument one in `error_handler`, so the vtable layout changes for `error_handler` and every user-derived class. Source compatibility is preserved as requested, but binaries built against the old header and linked against a new shared library will misbehave. `CMakeLists.txt` (`PROJECT_VERSION` / `SOVERSION`) and `ChangeLog.md` are not touched in this PR. That needs a version bump and a ChangeLog entry, and I'd like the `SOVERSION` question settled before merging since the Fedora packaging in `.distro` builds the shared library.


---

Heart-broken because _I_ think he is right about those 3 things and it took him only 1 minute to figure it out, while _I_ only saw the memory/reference-problem by intuition unable to grasp what would be the outcome.

The ABI-change needs to be handled but a middle-version increment - so that the .so is a link-target for existing binaries.

validation-error-copyness, well, needed I think.

Claude furthermore said:

### Review points not reflected in the code

- Three inline comments asking whether not, contains, and per-branch allOf should carry details["value"] when the subschema is a boolean, plus the reviewer's summary comment listing properties, dependencies, items, oneOf, and anyOf as missing the same, have no reply and no change. Those errors still ship an empty details object or only counts.
- The suggestion to normalize type to an array in details["value"] was not adopted. The raw keyword is stored, so a handler sees either a string or an array.
- The README says details are provided for not, which is technically true but the object is empty.

hughcars:
@pboettch all three are right, thanks. The reference wasn't really the bug though, the buffering was: anything the combinators hold has to own what it points at. So the second fix you listed can't work (the outer `anyOf` propagates after `object::validate` has returned), and the first leaves a struct that looks like a value but isn't.

Thinking a bit harder about your point I ended up with a redesign, see [`580cdfb`](https://github.com/pboettch/json-schema-validator/pull/384/commits/580cdfb). `allOf`/`anyOf`/`oneOf` first ask each branch whether it validates with a flag-only handler; only if the combination fails do they emit the summary and re-run the failed branches straight into the user's handler with the usual `case#` prefix. Nothing is stored, so nothing can dangle, and the throwing handler never re-runs anything. `validation_error` then becomes a plain value with the instance back as a parameter, `error(const validation_error &, const json &instance)`.

Side effects, all in the ChangeLog: checkers can run twice on the failure path with a non-throwing handler, the `allOf` summary sits at the combination's location like the other two, and `required`/`dependencies` stop copying their list into every error. 2.5.0 bump and ABI note included; `SOVERSION` unchanged as you asked, say if you'd rather I set it to 3.

@pjungkamp with this your example goes from 625 allocations on main (1003 here) to 408, the 500-entry enum case from 77k to 213. When I measured, valid documents and the throwing handler are level with or faster than main (the branch as pushed was ~35% slower). Collecting full diagnostics for a failure several combinations deep is ~2x main, same as this branch today; for yes/no callers there's now `is_valid()`, which is faster than main everywhere. Your `save()/restore()` and the `not`/`if`/`contains` rollback should sit on `580cdfb` unchanged.

src/json-validator.cpp
```diff
@@ -426,66 +498,67 @@ enum logical_combination_types {
...
+			if (!probe)
 				count++;
 			else {
 				patch.get_json().get_ref<nlohmann::json::array_t &>().resize(oldPatchSize);
-				esub.propagate(error_summary, "case#" + std::to_string(index) + "] ");
+				if (combine_logic == allOf) {
+					e.error(validation_error{ptr, "at least one subschema has failed, but all of them are required to validate", "allOf", {{"failed_subschema", index}}}, instance);
+					if (!probing())
+						report_branch(ptr, instance, e, index);
+					return;
```
pjungkamp, line 532: The current validation procedure is reporting errors for all branches in `allOf`. You should move this `return` into an `else` so we get a proper report for _all_ failed `allOf` branches when we're not probing.

#### Side Note

I still don't really understand why `allOf` is lumped together with `oneOf` and `anyOf`. The shared logic for default patch rollback and your "speculative evaluation" isn't needed for `allOf`; validation cannot succeed anymore once any branch has failed. I also don't think that there's any value in telling the user the branch of a failure in `allOf`. Both those behaviors only really make sense for `oneOf` and `anyOf`.

The validation of `allOf` should really just be a trivial loop:

```cpp
void validate(const json::json_pointer &ptr, const json &instance, json_patch &patch, error_handler &e) const final
{
  for (auto const &schema : subschemata_)
    schema->validate(ptr, instance, patch, e);
}
```

src/json-validator.cpp
```diff
@@ -426,66 +498,67 @@ enum logical_combination_types {
...
+				}
 			}
 
-			if (is_validate_complete(instance, ptr, e, esub, count, index))
+			if (combine_logic == anyOf && count == 1)
```
pjungkamp, line 536: This short-circuiting behavior for `anyOf` matches the behavior on the current `main` branch, so you can probably leave it as-is for this PR.

_But_ this behavior is not allowed according to JSON Schema Draft 7: [Section 3.3.2 _Annotations and Short-Circuit Validation_](https://json-schema.org/draft-07/draft-handrews-json-schema-validation-01#rfc.section.3.3.2) explicitly calls out that short-circuiting behavior is not allowed when processing annotations. The problem is that this JSON Schema validator uses the `default` annotation for building the returned default value patch. Short-circuiting on the first successful branch of `anyOf` means that the `default` values of subsequent branches that would have validated successfully are not being collected.

This early return is thus only correct when probing, not during ordinary validation.

pjungkamp (commented):
I'm not quite convinced by the probing scheme that you've implemented.

On a general note: We want to be able to return early and skip diagnostics when we're probing. The schema validator will currently continue validating past failures, which can lead to major inefficiencies when you're only checking the boolean validation outcome indicated by the first failure.

You're currently modifying the `error_handler` interface anyway, can't we make the new `error_handler::error` overload return a `bool` indicating the desire for an early return? 

Each `e.error(...)` site would just become `if (e.error(...)) return;`. All that "speculative evaluation" infrastructure could just be replaced with:

```cpp
class fail_fast_error_handler : public error_handler {
  bool failed;
public:
  bool error(...) final { failed = true; return true; }
  operator bool() const { return failed; }
};
```

I've typed out a simple POC in f365e290497ff4116fa5463f3e5672158bef867e. But I still see some more potential for optimization here.

<details>
<summary>More bpftrace experiments.</summary>

#### Schema

```json
{
  "type": "object",
  "additionalProperties": {
    "oneOf": [
      false,
      false,
      false,
      false,
      false,
      {
        "type": "object",
        "properties": {
          "type": {
            "type": "string",
            "const": "foo"
          },
          "bar": {
            "type": "string",
            "const": "baz"
          }
        },
        "additionalProperties": false
      },
      {
        "type": "object",
        "properties": {
          "type": {
            "type": "string",
            "const": "bar"
          },
          "baz": {
            "type": "string",
            "const": "bar"
          }
        },
        "additionalProperties": false
      },
      {
        "type": "object",
        "properties": {
          "type": {
            "type": "string",
            "const": "baz"
          },
          "foo": {
            "type": "string",
            "const": "baz"
          }
        },
        "additionalProperties": false
      }
    ]
  }
}
```

#### Instance

```
{
  "a": {
    "type": "foo",
    "bar": "baz"
  },
  "b": {
    "type": "bar",
    "baz": "bar"
  },
  "c": {
    "type": "baz",
    "foo": "baz"
  },
  "d": {
    "type": "foo",
    "bar": "baz"
  },
  "e": {
    "type": "bar",
    "baz": "bar"
  },
  "f": {
    "type": "baz",
    "foo": "baz"
  },
  "g": {
    "type": "foo",
    "bar": "baz"
  }
}
```

#### `main` (0a5c207a8a02dae8c781a205b7746395fbee1f56)

```
@target_alloc_count: 1276
@target_alloc_sizes:
[16, 32)             492 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
[32, 64)             434 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@       |
[64, 128)            308 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@                    |
[128, 256)            21 |@@                                                  |
[256, 512)             7 |                                                    |
[512, 1K)              7 |                                                    |
[1K, 2K)               7 |                                                    |
```

#### This branch (53c5372c3c705681b03b8542bac3aa7574864abb)

```
@target_alloc_count: 716
@target_alloc_sizes:
[16, 32)             380 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
[32, 64)             273 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@               |
[64, 128)             63 |@@@@@@@@                                            |
```

#### Fail-Fast (f365e290497ff4116fa5463f3e5672158bef867e)

```
@target_alloc_count: 562
@target_alloc_sizes:
[16, 32)             310 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
[32, 64)             203 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@                  |
[64, 128)             49 |@@@@@@@@                                            |
```

</details>

